PR #8, opened 2020-06-25
Add React.js support

alexcostaluiz:
Here I've followed the tutorials linked in the chat by Jesse:
https://docs.google.com/document/d/1GPy2zC1e9Rv0hBxM0K1dxk0tIIew4zx6Opr2zarYQo8/edit#
https://docs.google.com/document/d/1_DwUut7owgVkGYQXZO56Cj200wbUo5AiK917nOsYPlI/edit#heading=h.9qxyeilorj4t

The important things to note here are that the appengine application has been split into two services, one for React and one for Java. `dispatch.yaml` splits traffic to the site to the two different services based on requested URL. If the URL matches `/api/*` it will get sent to the Java service, otherwise traffic is directed to React.

The makefile has also been modified to accommodate the all the new and possible paths at which JS, HTML, and CSS files will be found.


alexcostaluiz:
Finally resolved all lint errors. Decided to remove CSS validation check from makefile and GitHub lint check because css-validator was not recognizing and throwing errors on valid property names and vendor extensions.

Also, updated the target Node.js version to `12.x.x`. Everyone will need to have Node.js version 12 installed to run a development server.

PR #11, opened 2020-06-29
Add authentication

alexcostaluiz:
Ready for review.

Created components Loading and Modal. Implemented Google Sign-In authentication in the Header component.


project/frontend/src/components/Loading.js
```diff
@@ -0,0 +1,17 @@
...
+import React from 'react';
+
+/**
+ * An animated loading icon.
+ */
+function Loading(props) {
+  return (
+    <div id='loading-comments' className='loading-ripple'>
+      <div></div>
+      <div></div>
```
kpancha, line 12: What are these divs for?

project/frontend/src/pages/Header.js
```diff
@@ -1,24 +1,105 @@
...
-          <button key='2' className='header-button'>
-            Log in
-          </button>,
-        ]
-      )}
-    </div>
-  );
+  let [GoogleAuth, setGoogleAuth] = useState(undefined);
+  let [currentUser, setCurrentUser] = useState(undefined);
+  let [signedIn, setSignedIn] = useState(false);
```
jesseelizabeth, line 11: Do we need to have both currentUser and signedIn? Can we just use currentUser.isSignedIn to determine whether or not the user is signed in?

project/frontend/src/pages/Header.js
```diff
@@ -1,24 +1,105 @@
...
+  let [isModalOpen, setIsModalOpen] = useState(false);
```
jesseelizabeth, line 12: I think these can all be const

project/frontend/src/pages/Header.js
```diff
@@ -1,24 +1,105 @@
...
-            Create Account
-          </button>,
-          <button key='2' className='header-button'>
-            Log in
-          </button>,
-        ]
-      )}
-    </div>
-  );
+  let [GoogleAuth, setGoogleAuth] = useState(undefined);
```
jesseelizabeth, line 9: Do we have to import GoogleAuth?

project/frontend/src/pages/Header.js
```diff
@@ -1,24 +1,105 @@
...
+      };
+    }
+  }, []);
+
+  if (currentUser === undefined) {
+    return (
+      <div id='header'>
+        <Loading />
+      </div>
+    );
```
jesseelizabeth, line 50: Will this just be stuck in loading until the user signs in?

jesseelizabeth:
Some overall thoughts on Header.js -
- Let's maybe pull out onGoogleAuthLoaded function (and any other auth related functionality you need) into another file called auth.js or something along those lines to make it clear this is for authentication related things. Then we can import those to header.
- Have another component SignIn.js that will just have the sign in related UI. Right now it looks like this is so minimal with just a 'Sign in with Google' button so this can maybe just stay in the main header component.
- Pull out all the signed in related things to another component. Header.js can just pass down the currentUser to this new component and then we can grab the profile, name, email, image etc. 
- This way in Header.js we can just have isSignedIn ? return <SignedInComponentName /> : return <SignInComponent> in the render part.

project/frontend/src/components/Loading.js
alexcostaluiz: The loading-ripple class targets the child divs in css and creates a little loading animation inside them.

alexcostaluiz: Removed the unnecessary id tag on the parent div.

project/frontend/src/pages/Header.js
alexcostaluiz: No I don't think we need both. The problem is that React re-renders a component when you update its state through the setters returned by `useState(...)` (e.g. `setSignedIn`, `setCurrentUser`), but will not actually re-paint a component if you update the state with an equal value. So when a user signs in, `setCurrentUser(...)` is called with the new user object, but the Google API uses the same object in memory and just overwrites the internal pointers, so react doesn't think the object has changed and then doesn't re-paint the component after re-rendering.

React uses `Object.is(...)` to test state equality which doesn't account for change to internal values inside an object, just checks the memory location. I tried cloning the user object when `setCurrentUser(...)` is called but this did not work, cloning objects in JavaScript turns out to be a lot more difficult that I would've expected. I also tried adding a new property with a random value to the user object every time `setCurrentUser(...)` is called, but of course again didn't work since the memory location is being compared, not the properties of the object.

My workaround to this problem is defining the `signedIn` state which is just a boolean instead of an object. Now when the user changes both `setCurrentUser(...)` and `setSignedIn(...)` are called with new values and `Object.is(...)` recognizes `false` and `true` as not being equal so React fully re-renders and re-paints the component.

project/frontend/src/pages/Header.js
alexcostaluiz: oops, yes thanks!

project/frontend/src/pages/Header.js
alexcostaluiz: No, I just decided to call that state variable `GoogleAuth`. Perhaps I should not use a capital letter as first letter to make it more clear that it's a variable? `GoogleAuth` is what they were calling that object in the API documentation so I just followed that pattern.

alexcostaluiz: The API library is imported in `index.html` inside the `public/` directory. That is were the `window.gapi` object comes from.

project/frontend/src/pages/Header.js
alexcostaluiz: The API is kind of weird, that's how I expected it to work as well, but after `window.gapi.auth2.init(...)` is called the user object is always initialized to an object regardless of whether a user is signed in. And if a user _is_ signed in (from a previous session or something) then `window.gapi.auth2.init(...)` actually signs them in automatically.

So basically the loading icon only shows up in the time after `DOMContentLoaded` and before `window.gapi.auth2.init(...)` is called, which in my experience is a fraction of a second.

Then, if a user is not logged in, `currentUser` will be assigned to some object that looks like `{wn: undefined, en: undefined}` and then when a user logs in, that object has its internal properties filled in.

project/frontend/src/pages/Header.js
alexcostaluiz: This problem is still a problem after the refactor using a global context state and Authentication component. An optimal solution would be to have React recognize that the `currentUser` object has changed but I'm not sure how to do this.

alexcostaluiz:
> Some overall thoughts on Header.js -
> 
> * Let's maybe pull out onGoogleAuthLoaded function (and any other auth related functionality you need) into another file called auth.js or something along those lines to make it clear this is for authentication related things. Then we can import those to header.
> * Have another component SignIn.js that will just have the sign in related UI. Right now it looks like this is so minimal with just a 'Sign in with Google' button so this can maybe just stay in the main header component.
> * Pull out all the signed in related things to another component. Header.js can just pass down the currentUser to this new component and then we can grab the profile, name, email, image etc.
> * This way in Header.js we can just have `isSignedIn ? return <SignedInComponentName /> : return <SignInComponent>` in the render part.
> 
> To start - let's at least pull out the signed in stuff into another component and then this will look much cleaner.

Did some research on React context and the `useContext` hook and it is the perfect solution to the user state issue. I created an Authentication component that controls the `currentUser` and `GoogleAuth` state and initializes the Google Sign-In API. This component provides a React context, `AuthContext`, that can be listened to by any descendant component using the `useContext` hook or the `Component.contextType` API. These child components will re-render every time a new context is available from the Authentication component context provider.

project/frontend/src/components/Authentication.js
```diff
@@ -0,0 +1,54 @@
...
+
+function Authentication(props) {
+  const [GoogleAuth, setGoogleAuth] = useState(undefined);
+  const [currentUser, setCurrentUser] = useState(undefined);
+  const [signedIn, setSignedIn] = useState(false);
+
+  const context = {
+    GoogleAuth: { get: GoogleAuth, set: setGoogleAuth },
+    currentUser: { get: currentUser, set: setCurrentUser },
+    signedIn: { get: signedIn, set: setSignedIn },
```
jesseelizabeth, line 13: Instead of creating a context object here, can we just put this in `React.createContext({})` above?

project/frontend/src/pages/Header.js
```diff
@@ -1,24 +1,76 @@
...
+        alt='Large profile.'
+      />
+      <h4 className='user-name'>{props.user.getName()}</h4>
+      <h5 className='user-email'>{props.user.getEmail()}</h5>
+      <button>Profile</button>
+      <br />
+      <button className='sign-out' onClick={props.signOut}>
+        Sign Out
+      </button>
+    </Modal>
```
jesseelizabeth, line 28: Can we separate UserModal out into it's own component and then just import it at the top

project/frontend/src/pages/Header.js
```diff
@@ -1,24 +1,76 @@
...
   );
 }
 
+function Header(props) {
+  const context = useContext(AuthContext);
```
jesseelizabeth, line 33: Let's name this authContext to be clear (in case more contexts are used in the future).

project/frontend/src/pages/Header.js
```diff
@@ -1,24 +1,76 @@
...
+    headerContent = <Loading />;
+  } else if (context.currentUser.get.isSignedIn()) {
+    const toggleModal = () => {
+      setIsModalOpen((prev) => !prev);
+    };
+
+    const signOut = () => {
+      setIsModalOpen(false);
+      context.GoogleAuth.get.signOut();
+    };
```
jesseelizabeth, line 47: Let's pull these out and define them up top with context and the modal stuff.

project/frontend/src/components/Authentication.js
```diff
@@ -0,0 +1,54 @@
...
+  };
+
+  useEffect(() => {
+    const onGoogleAuthLoaded = () => {
+      window.gapi.load('auth2', () => {
+        const auth2 = window.gapi.auth2.init({
+          client_id:
+            '21423804760-e3goj1cdhg49ojdf780mcq92qgshbr4v.apps.googleusercontent.com',
```
jesseelizabeth, line 21: Let's put this in a gitignore file and import it as a variable

project/frontend/src/components/Authentication.js
```diff
@@ -0,0 +1,54 @@
...
+    if (window.gapi) {
+      onGoogleAuthLoaded();
+      return () => {};
+    } else {
+      window.addEventListener('DOMContentLoaded', onGoogleAuthLoaded);
+      return () => {
+        window.removeEventListener('DOMContentLoaded', onGoogleAuthLoaded);
+      };
+    }
+  }, []);
```
jesseelizabeth, line 45: You shouldn't need these event listeners. It looks like the useEffect hook happens after the render so the DOM will be loaded.

See 'Does useEffect run after every render?' in this section https://reactjs.org/docs/hooks-effect.html#example-using-hooks

project/frontend/src/components/Authentication.js
```diff
@@ -0,0 +1,54 @@
...
+
+  const context = {
+    GoogleAuth: { get: GoogleAuth, set: setGoogleAuth },
+    currentUser: { get: currentUser, set: setCurrentUser },
+    signedIn: { get: signedIn, set: setSignedIn },
+  };
+
+  useEffect(() => {
+    const onGoogleAuthLoaded = () => {
+      window.gapi.load('auth2', () => {
```
jesseelizabeth, line 18: Add to a try, catch block

project/frontend/src/components/Authentication.js
alexcostaluiz: I don't think we can because this context object needs to be recreated every time the `Authentication` component re-renders due to state changes.

Also the parameter passed to `React.createContext(...)` is apparently only given to components listening to that context that don't have a `Provider` above them in the tree. I've put the `Authentication` really high in the tree so I don't think there's any way a consumer of `AuthContext` wouldn't have `Authentication` above it.

project/frontend/src/pages/Header.js
alexcostaluiz: done!

project/frontend/src/components/Authentication.js
alexcostaluiz: To be honest I don't think it is possible that gapi is not defined by the time this effect is running. I am certain now that defer scripts are run before DOMContentLoaded and according to some testing, I don't think this effect will ever run before DOMContentLoaded.

If you still think I should use a try/catch though, let me know and I will add it.

project/frontend/src/components/Authentication.js
alexcostaluiz: Going to implement this .env file in a different PR.

project/frontend/src/components/UserModal.js
```diff
@@ -0,0 +1,30 @@
...
+
+import React from 'react';
+
+import Modal from '../components/Modal.js';
+
+function UserModal(props) {
+  return (
+    <Modal
+      open={props.isModalOpen}
+      onDismiss={props.toggleModal}
```
jesseelizabeth, line 11: nit: pull these variables out. So before return add `const {isModalOpen, toggleModal, user, signOut} = props` and then here and below you can just do `open={isModalOpen}` etc.

project/frontend/src/components/Modal.js
```diff
@@ -0,0 +1,49 @@
...
+  const backdropStyle = {
+    display: props.open ? 'block' : 'none',
+  };
+
+  const modalStyle = {
+    top: props.top,
+    right: props.right,
+    bottom: props.bottom,
+    left: props.left,
+  };
```
jesseelizabeth, line 35: nit: pull these variables out as well and then you can just do `const modalStyle = { top, right, bottom, left }`

project/frontend/src/components/Authentication.js
```diff
@@ -0,0 +1,40 @@
...
+    signIn: () => {},
+    signOut: () => {},
+  });
+  const [currentUser, setCurrentUser] = useState({ get: undefined });
+
+  useEffect(() => {
+    window.gapi.load('auth2', () => {
+      const auth2 = window.gapi.auth2.init({
+        client_id:
+          '21423804760-e3goj1cdhg49ojdf780mcq92qgshbr4v.apps.googleusercontent.com',
```
jesseelizabeth, line 30: Let's put this clientId in a gitignore file and import it as a variable

project/frontend/src/components/Authentication.js
```diff
@@ -0,0 +1,54 @@
...
+/**
+ * A high-order component responsible for initializing and maintaining authentication state
+ * within the application. Serves the AuthContext to child components.
+ */
+function Authentication(props) {
+  const [GoogleAuth, setGoogleAuth] = useState({
+    signIn: () => {},
+    signOut: () => {},
+  });
+  const [currentUser, setCurrentUser] = useState({ get: undefined });
```
jesseelizabeth, line 24: What does the `get` do?

Also let's make this null instead of undefined.

project/frontend/src/components/Authentication.js
jesseelizabeth: Sorry, commented this again below before I saw this!

Sometimes when api keys are added on github, the service like flags it.  So you probably want to submit the .env PR before you submit this one.

project/frontend/src/components/Authentication.js
alexcostaluiz: The get doesn't do anything. The benefit here is that by wrapping the GoogleUser object inside an object React recognizes the user as a new object and actually rerenders the context state. I just chose the property name get because it makes sense in the context of calling `authContext.currentUser.get` instead of something redundant like `authContext.currentUser.user`.

jesseelizabeth: Gotcha, thanks for explaining!

project/frontend/src/pages/Header.js
```diff
@@ -1,22 +1,50 @@
...
+
+import { AuthContext } from '../components/Authentication.js';
+import Loading from '../components/Loading.js';
+import UserModal from '../components/UserModal.js';
 
 function Header(props) {
+  const authContext = useContext(AuthContext);
+  const [isModalOpen, setIsModalOpen] = useState(false);
+
+  if (!authContext.currentUser.get) {
```
jesseelizabeth, line 13: nit: we can also pull this out `const user = authContext.currentUser.get`

PR #12, opened 2020-06-29
Start algorithm

kpancha:
Create utility class RestaurantScorer to score restaurant with weighted fields (the weights are hard coded right now but will be replaced by user input once this connects to frontend). Maps each restaurant to its score.


project/src/main/java/com/google/sps/data/RestaurantScorer.java
```diff
@@ -0,0 +1,31 @@
...
+      score += priceLevelWeight;
+    }
+    if (restaurant.getPlaceTypes().contains(preferredDiningExp)) {
+      score += diningExpWeight;
+    }
+    if (restaurantRating >= 3) {
+      score += restaurantRating / 5;
+    } else if (restaurantRating > 0) {
+      // Subtract more points from score for lower rating. This calculation will be improved.
+      score -= (0.6 - restaurantRating / 5);
```
jesseelizabeth, line 27: Can this just be an else (instead of else if)? Or can we change this to restaurantRating < 3?  I know that it won't enter the else if block if it's >= 3, but I don't like having overlapping cases here (i.e. if it's >= 3, it also is > 0).

project/src/main/java/com/google/sps/data/RestaurantScorer.java
```diff
@@ -0,0 +1,34 @@
...
+    double score = 0;
+    double restaurantRating = restaurant.getAvgRating();
+    if (restaurant.getPriceLevel() == preferredPriceLevel) {
+      score += priceLevelWeight;
+    }
+    if (restaurant.getPlaceTypes().contains(preferredDiningExp)) {
+      score += diningExpWeight;
+    }
+    if (restaurantRating >= 3) {
+      score += restaurantRating / 5;
```
surbhimadan1995, line 26: Lets pull 5 & 3 into private static final class constants (for naming descriptiveness) 
MAX_RATING = 5 and AVERAGE_RATING = 3 as what they represent ?

project/src/main/java/com/google/sps/data/RestaurantScorer.java
kpancha: restaurantRating is set to -1 if there is no rating in the data that comes back from the API, so I had this to make sure that it only considers ratings that exist in the data, but maybe I can change the default value for the rating to something different that would work here?

surbhimadan1995: how about : 
boolean hasRating = restaurantRating != -1;
if (hasRating) {
   if (rating >= avg_rating) {
     score = ...
   } else {
     score = ... 
   }
}

its a little bit more nesting so optionally to avoid that, you can also return early: 
if (!hasRating): return score/maxPoints;
 if (rating >= avg_rating) {
     score = ...
 } else {
     score = ... 
 }

kpancha: I changed it to only have one if statement (if rating > 0) in commit 5033396 that bascially does the same thing (except adds slightly less points for a good rating)... let me know if this looks better. I can also change what I have in that commit to the hasRating way that Surbhi wrote above since that suggestion seems clearer to read in my opinion.

project/src/main/java/com/google/sps/servlets/RecommendationServlet.java
```diff
@@ -26,35 +28,49 @@
...
-    Set<Restaurant> restaurantSet = new HashSet<>();
+    restaurantSet = new HashSet<>();
+    restaurantScores = new HashMap<>();
     BufferedReader reader = request.getReader();
     String body = reader.readLine();
     try {
-      addRestaurantsToSet(body, restaurantSet);
+      JSONObject reqBody = new JSONObject(body);
+      JSONArray restaurantList = reqBody.getJSONArray("restaurants");
+      addRestaurantsToSet(restaurantList);
```
surbhimadan1995, line 43: I dont think we need the intermediate restaurantSet here ? 
scoreRestaurants can incorportate that I think:
`private static void scoreRestaurants(JSONArray restaurantList, JSONObject preferences) throws JSONException {
    for (int i = 0; i < restaurantList.length(); i++) {
      Restaurant restaurant = JsonToRestaurantParser.toRestaurant(restaurantList.getJSONObject(i));
      restaurantScores.put(restaurant, RestaurantScorer.score(restaurant, preferences));
    }
  }`

project/src/main/java/com/google/sps/servlets/RecommendationServlet.java
```diff
@@ -26,35 +28,49 @@
...
+      JSONObject preferences = reqBody.getJSONObject("preferences");
+      scoreRestaurants(preferences);
```
surbhimadan1995, line 45: and you can probably omit some local vars here as I think this is is clear enough even written as 
scoreRestaurants(reqBody.getJSONArray("restaurants"), reqBody.getJsonObject("preferences"));

project/src/main/java/com/google/sps/data/RestaurantScorer.java
```diff
@@ -0,0 +1,36 @@
...
+      score += priceLevelWeight;
+    }
+    if (restaurant.getPlaceTypes().contains(preferredDiningExp)) {
+      score += diningExpWeight;
+    }
+    // This statement executes if a rating exists for the resturant.
+    // Adds points to the score if the rating is > 3, subtracts points if rating is < 3.
+    if (restaurantRating > 0) {
+      score += (restaurantRating / MAX_RATING) - (MID_RATING / MAX_RATING);
+    }
```
surbhimadan1995, line 31: much cleaner - thanks 👍 ! 
should this be >= 0 though? I guess rating 0 is valid rating right since we represent no valid rating by -1?

kpancha: Looks like the rating on the API is on a scale of 1-5, so there would be no valid rating < 1: https://developers.google.com/places/web-service/details#fields "`rating` contains the place's rating, from 1.0 to 5.0, based on aggregated user reviews."

project/src/main/java/com/google/sps/data/RestaurantScorer.java
```diff
@@ -0,0 +1,36 @@
...
+    double score = 0;
+    double restaurantRating = restaurant.getAvgRating();
+    if (restaurant.getPriceLevel() == preferredPriceLevel) {
+      score += priceLevelWeight;
+    }
+    if (restaurant.getPlaceTypes().contains(preferredDiningExp)) {
+      score += diningExpWeight;
+    }
+    // This statement executes if a rating exists for the resturant.
+    // Adds points to the score if the rating is > 3, subtracts points if rating is < 3.
```
surbhimadan1995, line 28: nit: update comment:
// Adds points to the score if the rating is > MID_RATING, subtracts points if rating is < MID_RATING.

project/src/main/java/com/google/sps/servlets/RecommendationServlet.java
```diff
@@ -26,35 +28,49 @@
...
   private final Gson gson = new Gson();
   private static final Logger LOGGER = Logger.getLogger(RecommendationServlet.class.getName());
+  private static Set<Restaurant> restaurantSet;
+  private static Map<Restaurant, Double> restaurantScores;
 
   @Override
   public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
-    Set<Restaurant> restaurantSet = new HashSet<>();
+    restaurantSet = new HashSet<>();
+    restaurantScores = new HashMap<>();
```
jesseelizabeth, line 37: We can probably just initialize this above `private static Map<Restaurant, Double> restaurantScores = new HashMap<>();`

project/src/main/java/com/google/sps/data/RestaurantScorer.java
surbhimadan1995: sounds good 👍

kpancha: So I suppose I could represent "no valid rating" with a 0-- I just had set all fields that didn't exist to -1 by default but could make it 0 if that would make more sense?

project/src/main/java/com/google/sps/servlets/RecommendationServlet.java
kpancha: I think we want to initialize it to empty every time doPost is called or else it doesn't reset if the user enters different inputs and doPost gets called again, if that makes sense.

surbhimadan1995: Lets make it a local variable then -- I think one issue with declaring it static is that it becomes a shared instance across all RecommendationServlet objects that are instantiated, based on https://javaee.github.io/javaee-spec/javadocs/javax/servlet/http/HttpServlet.html   it looks like there's no reason for multiple instances of a servlet to exist as it seems to be tied to the lifecycle of the server itself , but if this is something that we need to re-create per-post request, its easier to avoid unintended sideeffects by keeping it scoped within doPost

project/src/main/java/com/google/sps/servlets/RecommendationServlet.java
```diff
@@ -29,32 +29,43 @@
...
+    // TODO: handle case where list of restaurants is empty.
     try {
-      addRestaurantsToSet(body, restaurantSet);
+      JSONObject reqBody = new JSONObject(body);
+      restaurantScores = scoreRestaurants(
+          reqBody.getJSONArray("restaurants"), reqBody.getJSONObject("preferences"));
     } catch (JSONException e) {
       LOGGER.log(Level.WARNING, "Error parsing JSON: " + e.getMessage());
+      restaurantScores = new HashMap<>();
     }
```
surbhimadan1995, line 43: Is this needed? now that we're handling the exception inside scoreRestaurants and it doesnt throw it anymore

kpancha: I believe line 37 could still throw an exception

surbhimadan1995: ah, that would be a case where reqBody itself isnt usable so we cant do anything more after that -- I think lets make this: 
```
try {
    JSONObject reqBody = new JSONObject(body);
    Map<Restaurant, Double> restaurantScores = scoreRestaurants(
          reqBody.getJSONArray("restaurants"), reqBody.getJSONObject("preferences"));
     // TODO: handle empty restaurant set, maybe by response.println("no results found"):
     // Sort restaurant entries by highest score and write list of entries to the response.
    List<Map.Entry<Restaurant, Double>> sortedRestaurants =
        new ArrayList(restaurantScores.entrySet());
    sortedRestaurants.sort(Map.Entry.comparingByValue(Collections.reverseOrder()));
    response.getWriter().println(gson.toJson(sortedRestaurants));
} catch (JSONException e) {
  log("Error parsing request body: " + e.getMessage());
  // TODO: print user-friendly output to response
  // such as response.getWriter().println("error processing request");
}
```

surbhimadan1995: in other words -- do the remaining processing in the try block as if we catch an Exception while parsing the request body, nothing to do except showing the user some appropriate error message

project/src/main/java/com/google/sps/servlets/RecommendationServlet.java
```diff
@@ -32,22 +32,21 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+      // Sort restaurant entries by highest score and write list of entries to the response.
+      List<Map.Entry<Restaurant, Double>> sortedRestaurants =
+          new ArrayList(restaurantScores.entrySet());
+      sortedRestaurants.sort(Map.Entry.comparingByValue(Collections.reverseOrder()));
+      response.getWriter().println(gson.toJson(sortedRestaurants));
     } catch (JSONException e) {
       LOGGER.log(Level.WARNING, "Error parsing JSON: " + e.getMessage());
-      restaurantScores = new HashMap<>();
+      // TODO: handle case where list of restaurants is empty.
+      response.getWriter.println();
```
kpancha, line 48: fix error

PR #13, opened 2020-06-29
Add distance calculation

kpancha:
New function in RestaurantScorer calculates the distance between two lat lng points and adds to the restaurant's score if the distance is within the specified radius.


project/src/main/java/com/google/sps/data/RestaurantScorer.java
```diff
@@ -1,36 +1,60 @@
 package com.google.sps.data;
 
+import com.google.gson.Gson;
+import java.util.HashMap;
+import java.util.Map;
 import org.json.JSONException;
 import org.json.JSONObject;
 
 public final class RestaurantScorer {
+  private static final double EARTH_RADIUS = 6371000; // meters
```
jesseelizabeth, line 10: Maybe we name this EARTH_RADIUS_METERS so it's clear this is in meters and then maybe we don't need the comment

project/backend/src/main/java/com/google/sps/data/RestaurantScorer.java
```diff
@@ -14,21 +18,33 @@ private RestaurantScorer() {}
...
   public static double score(Restaurant restaurant, JSONObject preferences) throws JSONException {
-    int preferredPriceLevel = preferences.getJSONObject("priceLevel").getInt("pref");
-    String preferredDiningExp = preferences.getJSONObject("diningExp").getString("pref");
     int priceLevelWeight = preferences.getJSONObject("priceLevel").getInt("weight");
     int diningExpWeight = preferences.getJSONObject("diningExp").getInt("weight");
-    int maxPoints = priceLevelWeight + diningExpWeight + 1;
+    int radiusWeight = preferences.getJSONObject("radius").getInt("weight");
+    int maxPoints = priceLevelWeight + diningExpWeight + radiusWeight + 1;
+    Map<String, Double> currLocation =
+        new Gson().fromJson(preferences.getJSONObject("currLocation").toString(), HashMap.class);
```
surbhimadan1995, line 26: Lets just use one static final instance of Gson: https://javadoc.io/doc/com.google.code.gson/gson/latest/com.google.gson/com/google/gson/Gson.html it says its thread-safe so we can safely reuse an instance

project/backend/src/main/java/com/google/sps/data/RestaurantScorer.java
```diff
@@ -14,21 +18,33 @@ private RestaurantScorer() {}
 
   /** Returns the percent match for a restaurant to a set of user preferences. */
   public static double score(Restaurant restaurant, JSONObject preferences) throws JSONException {
-    int preferredPriceLevel = preferences.getJSONObject("priceLevel").getInt("pref");
-    String preferredDiningExp = preferences.getJSONObject("diningExp").getString("pref");
     int priceLevelWeight = preferences.getJSONObject("priceLevel").getInt("weight");
     int diningExpWeight = preferences.getJSONObject("diningExp").getInt("weight");
-    int maxPoints = priceLevelWeight + diningExpWeight + 1;
+    int radiusWeight = preferences.getJSONObject("radius").getInt("weight");
+    int maxPoints = priceLevelWeight + diningExpWeight + radiusWeight + 1;
```
surbhimadan1995, line 24: add a comment here for why we're adding 1 at the end

project/backend/src/main/java/com/google/sps/data/RestaurantScorer.java
```diff
@@ -37,11 +53,29 @@ public static double score(Restaurant restaurant, JSONObject preferences) throws
     return percentMatch;
   }
 
+  /** Calculates the distance in meters between two sets of lat long coordinates. */
```
surbhimadan1995, line 56: nit: add "....... lat long coordinates using the Haversine formula." 

(useful to add it to comments for future readers to read more about it :) )

PR #15, opened 2020-06-30
First iteration of confidence score

kpancha:
- Skews ratings so that ratings with fewer ratings are weighed less.
- Starts off with NUM_INITIAL_RATINGS "fake ratings" that are worth the MID_RATING value.
- So, the less ratings there are, the more they will be skewed towards MID_RATING.
- Since restaurants with exactly MID_RATING do not affect the score, this effectively gives a lower weight to restaurants with fewer ratings.

Biggest area for improvement: NUM_INITIAL_RATINGS is completely arbitrary right now. It should ideally be based on the average number of ratings that each place we are looking at has (it's value should be less than the average but still based on the average). *note that this number does not actually represent ratings that exist

### DIFF
--- a/project/backend/src/main/java/com/google/sps/data/RestaurantScorer.java
+++ b/project/backend/src/main/java/com/google/sps/data/RestaurantScorer.java
@@ -6,6 +6,7 @@ import org.json.JSONObject;
 public final class RestaurantScorer {
   private static final int MAX_RATING = 5;
   private static final int MID_RATING = 3;
+  private static final int NUM_INITIAL_RATINGS = 20;
   /** Class should not be instantiated. */
   private RestaurantScorer() {}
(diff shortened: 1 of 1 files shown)


project/backend/src/main/java/com/google/sps/data/RestaurantScorer.java
```diff
@@ -6,6 +6,7 @@
 public final class RestaurantScorer {
   private static final int MAX_RATING = 5;
   private static final int MID_RATING = 3;
+  private static final int NUM_INITIAL_RATINGS = 20;
```
surbhimadan1995, line 9: add a comment here explaining the purpose of this var (& why its made up rn) - basically what you wrote in the PR description

PR #16, opened 2020-07-01
Add error message for JSON parsing

kpancha:


project/backend/src/main/java/com/google/sps/servlets/RecommendationServlet.java
```diff
@@ -39,11 +38,12 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
           new ArrayList(restaurantScores.entrySet());
       sortedRestaurants.sort(Map.Entry.comparingByValue(Collections.reverseOrder()));
+      response.setContentType("application/json");
       response.getWriter().println(gson.toJson(sortedRestaurants));
     } catch (JSONException e) {
       LOGGER.log(Level.WARNING, "Error parsing JSON: " + e.getMessage());
-      // TODO: handle case where list of restaurants is empty.
-      response.getWriter().println();
+      response.setContentType("text/html");
+      response.getWriter().println("Oops! We could not process your request at this time.");
```
surbhimadan1995, line 46: hmm HttpServletResponse has a sendError method we should probably use: https://javaee.github.io/javaee-spec/javadocs/javax/servlet/http/HttpServletResponse.html#sendError-int-java.lang.String- 
I think this would be SC_INTERNAL_SERVER_ERROR https://javaee.github.io/javaee-spec/javadocs/javax/servlet/http/HttpServletResponse.html#SC_INTERNAL_SERVER_ERROR

kpancha: Would I still want to log the error with Logger?

surbhimadan1995: I think we can remove that now that its been handled with sendError

PR #17, opened 2020-07-01
made a form with one input

miahsanchez:
I was having a lot of trouble trying to change everything from the old form at once so I decided to just start from the bottom up! This gives the option to only choose a cuisine right now and then I can build on this!


project/frontend/src/FindMatch/FindMatch.js
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+  // Updates the state of the input element so it holds the chosen value.
+  changeState(event) {
+    this.setState({ cuisine: event.target.value })
+  }
+
+  handleSubmit(event) {
+    alert("the cuisine you chose is: " + this.state.cuisine)
+  }
```
jesseelizabeth, line 19: Let's also add the event.preventDefault() at the end here so our page doesn't refresh.

project/frontend/src/FindMatch/FindMatch.js
```diff
@@ -0,0 +1,45 @@
...
+  constructor(props) {
+    super(props)
+    this.state = { cuisine: "" }
+    this.changeState = this.changeState.bind(this)
+    this.handleSubmit = this.handleSubmit.bind(this)
+  }
+
+  // Updates the state of the input element so it holds the chosen value.
+  changeState(event) {
+    this.setState({ cuisine: event.target.value })
```
jesseelizabeth, line 14: Fun trick here - you can just do `this.setState({ [event.target.name]: event.target.value })`. 
`event.target.name` will get the value of the `name` prop from the input (i.e. `name="cuisine"`). Just make sure you're naming the name props on your form inputs the same as you are naming your fields in state.
Once you get more inputs in your state this will be helpful, so you don't have to write out each one!

jesseelizabeth: How you have it is good for now because we only have cuisine in state, but in a follow up PR when you add another field to state, I would do this so you dont have to write multiple handleChange methods :)

project/frontend/src/FindMatch/FindMatch.js
```diff
@@ -0,0 +1,45 @@
+import React from "react"
+import ReactDOM from "react-dom"
+
+class PreferenceForm extends React.Component {
+  constructor(props) {
+    super(props)
+    this.state = { cuisine: "" }
```
kpancha, line 7: I think you could probably store all the fields in the state (i.e. priceLevel, distance, etc.)? I know this PR is just for one input but just thinking for how it can work with all the inputs

kpancha: And then for your changeState function you could pass in another parameter or something indicating which part of the state to change... something like (not sure if this exactly would work but you get the idea)
`changeState(event, key) {
  this.setState({key: event.target.value})
}`

miahsanchez: yeah that's what's going to happen!

PR #18, opened 2020-07-01
Use average num ratings to calculate num initial ratings

kpancha:
Change from previous confidence score: uses the 25th percentile of number of ratings as the number of initial dummy ratings (again subject to change-- we might want to lower the percentile).

Uses a statistics class from Apache commons math called DescriptiveStatistics.


project/backend/src/main/java/com/google/sps/data/RestaurantScorer.java
```diff
@@ -31,16 +34,38 @@ public static double score(Restaurant restaurant, JSONObject preferences) throws
...
+  public static DescriptiveStatistics createDescriptiveStatistics(JSONArray restaurantList) {
+    DescriptiveStatistics statistics = new DescriptiveStatistics();
+    for (int i = 0; i < restaurantList.length(); i++) {
+      try {
+        // int numRatings = restaurantList.getJSONObject(i).has("user_ratings_total") ?
+        // restaurantList.getJSONObject(i).getInt("user_ratings_total") : 0;
+        // statistics.addValue(numRatings);
+        statistics.addValue(restaurantList.getJSONObject(i).getInt("user_ratings_total"));
+      } catch (JSONException e) {
+        // LOGGER.log("Error parsing JSON: " + e.getMessage());
```
surbhimadan1995, line 56: remove commented out lines?

kpancha: ohh I meant to leave a comment about this-- I was just wondering which would be a better way to code this. The commented out lines are an alternative way (and then I wouldn't keep lines 54 and 57) to do the same thing. Does one look better than the other?

surbhimadan1995: you'd still want to keep addValue(0) in the catch block right? or does the commented out way mean the exception wont need to be handled?

kpancha: 51-52 would cover that because it checks to see if user_ratings_total is a field and if it is not, then sets numRatings to 0. So the catch block would only execute in the case of an unexpected error.

surbhimadan1995: I see - i think we need addValue(0) in either case then : when user rating is not a field and when there';s an unexpected error so:
```
try {
  JSONObject restaurantObj =  restaurantList.getJSONObject(i);
  int numRatings = restaurantObj.has("user_ratings_total") ?restaurantObj.getInt("user_ratings_total") : 0;
  statistics.addValue(numRatings);
} catch (JSONException e) {
    statistics.addValue(0);
}
```

PR #19, opened 2020-07-06
Implement restaurant card

alexcostaluiz:
Ready for review.

Implements a restaurant card that will be used throughout the app. The RestaurantCard component is also composed of the ImageSlider and RatingStars components.

Checkout the card interactively [here](https://mak-step-2020.appspot.com/card).

Screenshot 2020-07-07 at 7 46 55 PM


project/frontend/src/index.js
```diff
@@ -10,18 +10,26 @@ import Header from './pages/Header.js';
...
+
 import * as serviceWorker from './serviceWorker';
 
 ReactDOM.render(
   <React.StrictMode>
     <Authentication>
       <Router>
         <Switch>
+          
+          <Route path='/test'>
```
miahsanchez, line 23: is this the line that gets the card to show up on the /test URL ?

alexcostaluiz: Yeah, whatever children are added under a `Route` component will be the ones visible on that page. It is kind a false page though, when you enter a new url the page doesn't actually reload it just edits the DOM. So when you go from '/' to '/test', React replaces the children under `<Route path='/'>` with <RestaurantCard />.

project/frontend/src/components/ImageSlider.js
```diff
@@ -0,0 +1,223 @@
...
+      const newLeft = newSlide * -width;
+      return {
+        left: newLeft,
+        lastLeft: newLeft,
+        slide: newSlide,
+        slideCount: state.slideCount,
+        shifting: true,
+      };
+    }
+    default: {
```
surbhimadan1995, line 134: for the last 4 cases atleast -- could we just pull the return out to the end of the method and just break in the individual cases since they're the same vars ? (might need declaring the vars newLeft etc before the switch)

surbhimadan1995:
The card looks amazing!! :D

project/frontend/src/components/ImageSlider.js
alexcostaluiz: Done! Was able to remove a lot of redundancy by pulling out the return statement!

project/frontend/src/components/ImageSlider.js
```diff
@@ -0,0 +1,186 @@
...
+  PREV_SLIDE: 3,
+  NEXT_SLIDE: 4,
+};
+
+/**
+ * The default width of a slide.
+ *
+ * @type {number}
+ */
+const width = 344;
```
jesseelizabeth, line 26: Does this scale with the width of the screen?

project/frontend/src/index.js
```diff
@@ -13,18 +13,27 @@ import MapContainer from './components/ResultsMap.js';
...
+
 ReactDOM.render(
   <React.StrictMode>
     <Authentication>
       <Router>
         <Header />
         <Switch>
-          <Route exact path='/' component={HomePage} />
+          <Route path='/card'>
+            <RestaurantCard restaurant={restaurant} details={details} />
```
jesseelizabeth, line 30: Do we not need component={RestaurantCard}?

alexcostaluiz: No, you can either specify the child in `component=` or just give the `Route` explicit children. I needed to give props to `RestaurantCard` so I went the descendant route instead of `component=`.

project/frontend/src/components/ImageSlider.js
alexcostaluiz: No, right now the card dimensions are hard-coded. I do think it is necessary to scale the card based on different screen sizes but I'd like to make those changes in a future commit, as scaling the card sizes will have a lot to do with the CardDeck and CardStack components as well.

jesseelizabeth: Ok, lets leave a TODO here to come back and do that!

project/frontend/src/components/ImageSlider.js
```diff
@@ -0,0 +1,186 @@
...
+  }
+
+  if (action.type !== Action.SWIPING) {
+    left = slide * -width;
+    lastLeft = left;
+    shifting = true;
+  }
+
+  return { left, lastLeft, slide, slideCount, shifting };
+}
```
jesseelizabeth, line 101: nit: (for a follow up CL) move the reducer logic out into it's own file ImageSliderReducer.js and put all your reducers in their own directory

alexcostaluiz: I'll add a todo.

project/frontend/src/components/SampleRestaurant.js
```diff
@@ -0,0 +1,360 @@
...
+    address: '2162 Larkin St, San Francisco, CA 94109, USA',
+    avgRating: 4.6,
+    id: 'apr75h4bni2pf98h4inujnksjrliu34',
+    latLngCoords: {
+      lat: 37.7979601,
+      lng: -122.4203648,
+    },
+    name: 'Amarena',
+    numRatings: 201,
+    placeTypes: ['Italian', 'Restaurant'],
```
kpancha, line 17: Doesn't look like this would affect anything, but right now cuisine type isn't a part of placeTypes. The possible strings in the array are 'meal_delivery', 'meal_takeaway', and 'restaurant'

project/frontend/src/index.js
```diff
@@ -13,18 +13,27 @@ import MapContainer from './components/ResultsMap.js';
...
+          </Route>
+
           <Route path='/find-match' component={PreferenceForm} />
           <Route path='/match-results' component={MatchResults} />
           <Route path='/mapRoute' component={MapContainer} />
+          <Route path='/' component={HomePage} />
```
kpancha, line 36: Did you move this to the bottom and get rid of "exact" so that any other url directs here? If so, I think leaving the exact path above and then using redirect at the bottom might be better but not sure https://reactrouter.com/web/api/Redirect

alexcostaluiz: I did. Yea I like your idea much better, I will change it.

alexcostaluiz: So using the Redirect causes some problems with the Authentication component for some reason, so I'm going to postpone this change for another PR, because I anticipate having to make some changes to Authentication.

alexcostaluiz: I will leave a TODO for now.

PR #20, opened 2020-07-06
Multiple cuisine types

kpancha:
Modify script.js to handle multiple cuisine types in an array. Creates array of promises (one fetch call for each cuisine type) and then handles all the results together. Currently, the second cuisine type is hard coded in.


project/backend/src/main/webapp/script.js
```diff
@@ -14,7 +14,9 @@
 
 /* exported getRecommendation */
 function getRecommendation() {
-  const cuisineType = document.getElementById('cuisine').value;
+  const cuisineTypes = [];
+  cuisineTypes.push(document.getElementById('cuisine').value);
+  cuisineTypes.push('chinese');
```
surbhimadan1995, line 19: do we still need this hardcoding?

kpancha: removed

PR #21, opened 2020-07-06
Multiple inputs

miahsanchez:
Added distance, dining experience, price, longitude, latitude, and an open option!


project/frontend/src/FindMatch/FindMatch.js
```diff
@@ -0,0 +1,129 @@
...
+          What cuisine?
+          <select
+            name='cuisine'
+            id='cuisine'
+            onChange={this.changeState}
+            value={this.state.cuisine}>
+            {cuisines.map((cuisine) => (
+              <option value={cuisine}>
+                {cuisine[0].toUpperCase() + cuisine.slice(1)}
+              </option>
```
kpancha, line 52: You can just do the proper capitalization in the array. I realized since we are just using this for text search, case doesn't matter so I would just capitalize the first letter of each word in 32

miahsanchez: just changed it!

project/frontend/src/FindMatch/FindMatch.js
```diff
@@ -0,0 +1,129 @@
...
+        </label>
+        <label>
+          Price Level
+          <select
+            name='price_level'
+            id='price_level'
+            onChange={this.changeState}
+            value={this.state.price_level}>
+            {prices.map((price) => (
+              <option value={price}>{price}</option>
```
kpancha, line 91: Let's have the values for price match the numeric scale that the API uses. so like {'Low': 1, 'Medium': 2, 'High': 3}. There is also a 0 and 4, for free and very expensive, respectively, but I don't think 0 is necessary. Not sure if we want to try to include price level 4 as well...

See price_level here for more: https://developers.google.com/places/web-service/details#fields

kpancha:
I think we should have the values for each field blank by default (maybe we could have a placeholder that says "select" or something). Entering a location (lat and lng) should be mandatory to submit the form (this is needed for the API call) so I think for this we can use the required attribute https://stackoverflow.com/questions/52902613/how-can-i-add-required-attribute-to-the-fileds-in-react-js/52902707 . 

So we can give the user the option to choose what fields they enter instead of requiring the user to select a value for all fields. We can work on this in a follow up PR (we'll have to modify script.js a little for this too), but just some food for thought :)

project/frontend/src/FindMatch/FindMatch.js
```diff
@@ -0,0 +1,129 @@
...
+
+  // Updates the state of the input element so it holds the chosen value.
+  changeState(event) {
+    const value =
+      event.target.name === 'open' ? event.target.checked : event.target.value;
+    this.setState({ [event.target.name]: value });
+  }
+
+  handleSubmit(event) {
+    alert('the cuisine you chose is: ' + this.state.cuisine);
```
surbhimadan1995, line 27: is this a pop-up alert? I wonder if this'll look too intrusive (and if we need an alert like this at all?)

project/frontend/src/FindMatch/FindMatch.js
```diff
@@ -0,0 +1,131 @@
...
+    event.preventDefault();
+  }
+
+  render() {
+    const cuisines = ['Italian', 'Mexican', 'Indian'];
+    const distances = [1, 5, 10, 25];
```
surbhimadan1995, line 33: lets name this distances_in_miles to make the units explicit

project/frontend/src/FindMatch/FindMatch.js
surbhimadan1995: Lets include level 4 here, maybe add "Very High" : 4 ?

project/frontend/src/FindMatch/FindMatch.js
miahsanchez: Jesse told me to just add this for now as a filler(until we do more with the handleSubmit function)

surbhimadan1995: 👍  can you add a TODO here to address this in follow ups then?

project/frontend/src/FindMatch/FindMatch.js
```diff
@@ -0,0 +1,132 @@
...
+
+  handleSubmit(event) {
+    // TODO: Remove alert and implement.
+    alert('the cuisine you chose is: ' + this.state.cuisine);
+    event.preventDefault();
+  }
+
+  render() {
+    const cuisines = ['Italian', 'Mexican', 'Indian'];
+    const distances_on_miles = [1, 5, 10, 25];
```
surbhimadan1995, line 34: nit: in instead of on

PR #24, opened 2020-07-07
Change Map's center coordinates

miahsanchez:
I added the changeCenter function which changes the lat lng coordinates to something random. I checked that it works by adding a button to ensure the coordiantes change when the button was clicked but I deleted the button! Added some TODOs that I plan on working on in my next PR

### DIFF
--- a/project/frontend/src/ResultsMap.js
+++ b/project/frontend/src/ResultsMap.js
@@ -21,15 +21,13 @@ export class MapContainer extends React.Component {
   }
 
   render() {
+    const { lat, lng } = this.state;
     return (
       <Map
         google={this.props.google}
         zoom={14}
         style={mapStyle}
-        center={{
-          lat: this.state.lat,
-          lng: this.state.lng,
-        }}
+        center={(lat, lng)}
       />
     );
   }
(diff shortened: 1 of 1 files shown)


project/frontend/src/ResultsMap.js
```diff
@@ -8,15 +8,27 @@ const mapStyle = {
...
       <Map
         google={this.props.google}
         zoom={14}
         style={mapStyle}
-        initialCenter={{
-          lat: -1.2884,
-          lng: 36.8233,
+        center={{
+          lat: this.state.lat,
+          lng: this.state.lng,
```
jesseelizabeth, line 31: nit: after render pull these out `const { lat, lng } = this.state` and then here you can just do center = { lat, lng }

miahsanchez: quick question about this! if i add more states for my Map could I still do `const {lat,lng} = this.state` ? im adding more states in another branch and i'm wondering if it will affect this

jesseelizabeth: More fields to this state? Shouldn't affect that! Thats's just the equivalent of saying this.state.lat and this.state.lng

miahsanchez:
This function I wrote is no longer needed because once we get the restaurant list

PR #26, opened 2020-07-08
Add message if no results show up

kpancha:


project/frontend/src/FindMatch/FindMatch.js
```diff
@@ -165,6 +165,16 @@ class PreferenceForm extends React.Component {
           restaurants = restaurants.concat(restaurant.results);
         }
 
+        if (restaurants.length === 0) {
+          let message = "We can't find any restaurants near you";
+          if (this.state.open) {
+            message += ' that are currently open';
+          }
+          message += '. Please try changing your preferences.';
```
surbhimadan1995, line 173: nit: lets make this 
+= ' that match your preferences. Please try changing your preferences' 
(im just being pedantic about adding context to the cant find restaurants near you 🙃  )

project/frontend/src/FindMatch/FindMatch.js
```diff
@@ -165,6 +165,16 @@ class PreferenceForm extends React.Component {
...
+          alert(message);
```
surbhimadan1995, line 174: possibly for a follow up- but I still think a pop-up alert could be too jarring, maybe just displaying this message at the place where we would show the results if there were any?

project/backend/src/main/java/com/google/sps/data/RestaurantScorer.java
```diff
@@ -115,7 +115,7 @@ private static long roundMetersToMiles(double numMeters) {
   /** Skews ratings so that ratings with a lower number of ratings are weighed less. */
   private static double calculateRatingScore(
       double avgRating, int numRatings, DescriptiveStatistics statistics) {
-    int numInitialRatings = (int) Math.round(statistics.getPercentile(25));
+    int numInitialRatings = (int) Math.round(statistics.getPercentile(10));
```
surbhimadan1995, line 118: is this change related ?

project/frontend/src/FindMatch/FindMatch.js
jesseelizabeth: you could just create a NoResults component that would be rendered in this case

project/backend/src/main/java/com/google/sps/data/RestaurantScorer.java
kpancha: Oh no it's not, probably should've put this in a different PR

project/frontend/src/scripts/recommendation_script.js
```diff
@@ -8,6 +8,17 @@ export default function getRecommendation(preferences, callback) {
     )
     .then((data) => {
       let restaurants = [];
+      if (restaurants.length === 0) {
+        let message = "We can't find any restaurants near you";
+        if (this.state.open) {
+          message += ' that are currently open';
+        }
+        message +=
+          ' and match your preferences. Please try changing your preferences.';
```
surbhimadan1995, line 17: nit: "that" instead of "and"

PR #27, opened 2020-07-08
Implement restaurant card stack

alexcostaluiz:
Ready for review.

Sorry for the size of this PR, got a little carried away. Implemented both the `RestaurantCardDeck` and `RestaurantCardStack` components. Implementing these components also required some refactoring in the `RestaurantCard` and `ImageSlider` components.

Checkout a demo of the `RestaurantCardDeck` and `RestaurantCardStack` [here](https://mak-step-2020.appspot.com/deck) (`Deck` on the left and `Stack` on the right). Also I've realized that the deck/stack might not fit within the vertical space of a smaller screen (e.g. my pixelbook screen) so you may want to `ctrl + -` a few times to see the whole component as you click its buttons.


project/frontend/src/components/RestaurantCardStack.js
```diff
@@ -0,0 +1,89 @@
...
+ */
+function RestaurantCardStack(props) {
+  const [currentCardIndex, setCurrentCardIndex] = useState(0);
+
+  const maxScaleFactor = 1.0;
+  const minScaleFactor = 0.85;
+  const numberOfCollapsedCards = 3;
+  const uncollapsedCardHeight = 487;
+  const offsetBetweenCollapsedCards = 24;
+  const offsetBeforeUncollapsedCard = 152;
```
kpancha, line 21: Wondering if you'd want to make these relative to each other so that if you have to resize, the proportions will stay constant? Not fully sure how the whole design works but just a thought

project/frontend/src/components/RestaurantCardDeck.js
```diff
@@ -0,0 +1,127 @@
...
+  const minScaleFactor = 0.9;
+  const numberOfVisibleCards = 3;
+  const offsetBetweenCards = 24;
+  const transitionString =
+    'top 0.75s cubic-bezier(0.35, 0.91, 0.33, 0.97), ' +
+    'left 0.75s cubic-bezier(0.35, 0.91, 0.33, 0.97), ' +
+    'transform 0.75s cubic-bezier(0.35, 0.91, 0.33, 0.97), ' +
+    'opacity 0.75s cubic-bezier(0.35, 0.91, 0.33, 0.97)';
+  const cards = props.cards.map((card, index) => {
+    if (index < state.index - 7 || index > state.index + 7) {
```
surbhimadan1995, line 68: 7 == number of cards shown? lets use a named const for this

project/frontend/src/components/RestaurantCardDeck.js
```diff
@@ -0,0 +1,127 @@
...
+function RestaurantCardDeck(props) {
+  const [state, dispatch] = useReducer(reducer, {
+    index: 0,
+    history: new Array(props.cards.length),
+  });
+
+  const maxScaleFactor = 1.0;
+  const minScaleFactor = 0.9;
+  const numberOfVisibleCards = 3;
+  const offsetBetweenCards = 24;
```
surbhimadan1995, line 61: nit: is this px? if so, lets name it something like offsetPxBetweenCards

project/frontend/src/components/RestaurantCardDeck.js
```diff
@@ -0,0 +1,127 @@
...
+    case Action.INTERESTED:
+    case Action.NOT_INTERESTED:
+    case Action.NEUTRAL:
+      history[index] = action.type;
+      index++;
+      break;
+    default:
+      // Action.BACK
+      if (index > 0) index--;
+      break;
```
surbhimadan1995, line 40: nit: break is not needed at the last case of a switch

project/frontend/src/components/RestaurantCardDeck.js
```diff
@@ -0,0 +1,127 @@
...
+  const history = previous.history;
+
+  switch (action.type) {
+    case Action.INTERESTED:
+    case Action.NOT_INTERESTED:
+    case Action.NEUTRAL:
+      history[index] = action.type;
+      index++;
+      break;
+    default:
```
surbhimadan1995, line 37: Lets add the back action explicitly as a case as well and it can fall through to default 

case Action.BACK:
default:
     if index > 0: index--

project/frontend/src/index.js
```diff
@@ -9,25 +9,57 @@ import Authentication from './components/Authentication.js';
...
+import RestaurantCardDeck from './components/RestaurantCardDeck.js';
+import RestaurantCardStack from './components/RestaurantCardStack.js';
 
 import * as serviceWorker from './serviceWorker';
 
 // TODO: remove
-import { restaurant, details } from './components/SampleRestaurant.js';
+import generateRestaurant from './components/SampleRestaurant.js';
+const cards = [];
+for (let i = 0; i < 20; i++) {
```
surbhimadan1995, line 26: extract 20 to a named const as well

project/frontend/src/components/RestaurantCardStack.js
surbhimadan1995, line 21: same for adding units into the name like uncollapsedCardHeightPx etc

project/frontend/src/components/RestaurantCardStack.js
```diff
@@ -0,0 +1,89 @@
...
+    const offset =
+      relativeIndex === 0
+        ? offsetBeforeUncollapsedCard
+        : relativeIndex > 0
+        ? offsetBeforeUncollapsedCard +
+          uncollapsedCardHeight +
+          Math.min(relativeIndex, numberOfCollapsedCards) *
+            offsetBetweenCollapsedCards
+        : Math.max(numberOfCollapsedCards + relativeIndex, 0) *
+          offsetBetweenCollapsedCards;
```
surbhimadan1995, line 42: errr this is a bit of personal pref but I think chained ternaries especially given that the expressions within a clause aren't trivial here -- are a bit hard to follow while reading -- consider just using conditionals instead? 
```
const offset;
if ( relativeIndex === 0):
    offset = offsetBeforeUncollapsedCard;
else if (relativeIndex > 0):
    offset = offsetBeforeUncollapsedCard + uncollapsedCardHeight +
          Math.min(relativeIndex, numberOfCollapsedCards) * offsetBetweenCollapsedCards;
else:
   offset =  Math.max(numberOfCollapsedCards + relativeIndex, 0) * offsetBetweenCollapsedCards;
```

project/frontend/src/components/RestaurantCardStack.js
alexcostaluiz: I think this is a good idea, but I'd like to put it off until I implement everything around resizing the cards based on screen size because I'm not sure what that's going to look like yet and changes now  will probably just get overwritten when I actually do the whole refactor.

project/frontend/src/components/RestaurantCard.js
```diff
@@ -17,49 +17,110 @@ import Clock from '../assets/clock.svg';
...
-      '&maxwidth=344'
-  );
+  const photoUrls = details
+    ? details.result.photos.map(
+        (photo) =>
+          'https://maps.googleapis.com/maps/api/place/photo?photoreference=' +
+          photo.photo_reference +
+          '&key=' +
+          process.env.REACT_APP_GOOGLE_API_KEY +
+          '&maxwidth=344'
```
jesseelizabeth, line 35: nit: you can use a template literal here ``https://maps.googleapis.com/maps/api/place/photo?photoreference=${photo.photo_reference}&key=${process.env.REACT_APP_GOOGLE_API_KEY}&maxwidth=344``
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

project/frontend/src/components/RestaurantCard.js
```diff
@@ -17,49 +17,110 @@ import Clock from '../assets/clock.svg';
...
+  // rendered with their default full heights (style.height = '') regardless of whether
+  // `collapsed` is true or false. This allows the full height to be captured in the
+  // above layoutEffect before this component is painted to the screen. The layoutEffect
+  // also updates the collapsible elements height according to the value of `collapsed`
+  // before painting so there is no flicker even though an initially collapsed card will
+  // be first rendered at full height.
+  //
+  // This is all necessary to avoid hardcoding the maximum height of the collapsible
+  // elements, as their height may vary due to word wrapping within the collapsible
+  // elements (e.g. long addresses or website urls).
```
jesseelizabeth, line 71: Will the cards all be different heights? It may be best to have a standard size that all cards should be. And then you would just allocate a fixed amount of lines for address, urls etc. If it happens to be longer, you just truncate the text.

project/frontend/src/index.js
```diff
@@ -13,11 +13,30 @@ import HomePage from './pages/HomePage.js';
...
+    'apr75h4bni2pf98h4inujnksjrliu34' + i
+  );
+  if (i === 3) {
+    data.restaurant.key.name += ' Of the California State of the US';
+  } else if (i === 4) {
+    data.restaurant.key.address = '123 Main Street';
+  } else if (i === 7) {
+    data.details.result.website +=
+      '?somethingextra=123&antherthing=1234567890&type=json';
+  }
```
surbhimadan1995, line 37: does this logic mean that we see the address peeking out for card # 4 etc ? (I think this block could use comments explaining that -- the 3,4,7 seems arbitrary without explanation)

project/frontend/src/components/RestaurantCard.js
```diff
@@ -17,49 +17,106 @@ import Clock from '../assets/clock.svg';
...
+    collapsibleMaxHeight.current = collapsible.current.offsetHeight;
+    if (collapsed) {
+      collapsible.current.style.height = '0px';
+      restaurantName.current.classList.add('collapsed');
+    } else {
+      collapsible.current.style.height = collapsibleMaxHeight.current + 'px';
+    }
+    // We only want this effect to run once (the very first time this component
+    // is laid out) so that we may capture the uncollapsed height of the collapsible
+    // div within this card and save it to a persistent ref (collapsibleMaxHeight).
```
jesseelizabeth, line 49: This seems hacky, not sure if I'm fully understanding this. If we just want this to capture the uncollapsed height of the card, why do we have the if else?

alexcostaluiz: It's necessary to set the collapsible div to its true height according to the `collapsed` prop. On the first render of the component the collapsible div will be rendered at full height regardless of whether `collapsed` is true or false, so that height can be captured in this `useLayoutEffect` and then the height is updated to reflect the `collapsed` state right before the component is painted to the screen. `useLayoutEffect` runs right before components are painted.

alexcostaluiz: The overall reason for this method is that the height of the collapsible div will be different for different restaurant cards, and for it to be animated smoothly between collapsed and uncollapsed states, I need an explicit `px` value target. Going from `0px` to `auto` or 'null' will not animate, it will just instantly render at full height. This is a common problem faced with css transitions, and from what I've seen, calculating and saving the height of a variable height component is a common solution to the issue.

project/frontend/src/index.js
alexcostaluiz: This stuff was really just for testing purposes. I wanted to have the cards render with different collapsible div heights, so I changed the text content of some attributes. This will be removed in a future PR, if it was not already removed in #32.

PR #28, opened 2020-07-09
Add callback function to redirect to MatchResults

kpancha:
Adds a callback function to redirect to MatchResults after getRecommendation finishes.


project/frontend/src/pages/MatchResults.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+
+class MatchResults extends React.Component {
+  render() {
+    const { data } = this.props.location;
```
surbhimadan1995, line 5: is it possible to name this restaurantsList or something more descriptive than data?

PR #30, opened 2020-07-10
Adding Marker and InfoWindow 

miahsanchez:
Screen Shot 2020-07-10 at 2 50 19 PM
Marker automatically shows up and the infoWindow shows up when you click on the marker. there is also an option to close out the infoWindow.


project/frontend/src/components/ResultsMap.js
```diff
@@ -8,15 +8,45 @@ const mapStyle = {
 };
 
 function MapContainer(props) {
+  // Declare all the states for this functional component
```
jesseelizabeth, line 11: This comment is probably unnecessary

PR #31, opened 2020-07-10
Make PreferenceForm component a popup modal over HomePage

alexcostaluiz:
Ready for review!

Adds the `PreferenceForm` component to the `HomePage` as a popup modal. Also configures the user traversal from `HomePage -> PreferenceForm -> MatchResultsPage`.

Also adds some utility to the Modal component, allowing it to be centered or centered horizontally using the `center` and `centerHorizontal` props.


project/frontend/src/pages/HomePage.js
```diff
@@ -1,21 +1,27 @@
...
 function HomePage() {
-  return (
-    <div className='container u-full-width'>
+  const [isPreferenceFormOpen, setIsPreferenceFormOpen] = useState(false);
+
+  const togglePreferenceForm = () => {
+    setIsPreferenceFormOpen((prev) => !prev);
+  };
+
+  return [
```
jesseelizabeth, line 15: Can you return an array like this in the render?

project/frontend/src/pages/MatchResultsPage.js
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
+
+function MatchResultsPage(props) {
+  console.log(props.location.state);
```
jesseelizabeth, line 4: remove console.log?

project/frontend/src/components/PreferenceForm.js
```diff
@@ -201,7 +191,9 @@ class PreferenceForm extends React.Component {
             onChange={this.changeState}
           />
         </label>
-        <button type='submit'>Submit</button>
+        <Link to={{ pathname: '/match-results', state: this.state }}>
```
jesseelizabeth, line 192: Is there a benefit to doing the form submit this way rather than the hadleSubmit function?

project/frontend/src/pages/HomePage.js
alexcostaluiz: Yes, apparently as of React 16.x.x you can return arrays instead of having to return a root node. It gets added to the DOM like you'd expect with the elements in the array being siblings. Alternatives are wrapping in an extra `div` or 'React.Fragment' component, but I think the array is okay and cuts down on nesting.

project/frontend/src/components/PreferenceForm.js
alexcostaluiz: We decided we want to send the form state to the new page and call `getRecommendations` in the `/match-results` page instead of within this form so that users don't press submit and then get stuck looking at their form for a second or two. So doing it this way, a user presses submit and goes straight to the `/match-results` page where a loading popup will render while we wait to get data from `getRecommendations`.

We could still do this using `handleSubmit` but instead of writing an extra event handler this `Link` component will do all the work for us. Also, using this `Link` makes sure we take advantage of React's one page application functionality wherein the user's browser wont actually refresh on submit and the page navigation is done "in-place" so there's no re-downloading content from the server.

jesseelizabeth: Does this still appropriately handle the required fields?

alexcostaluiz: Oh shoot it does not.

jesseelizabeth: Yeah I think the form onSubmit handles a lot of form related things under the hood. The event.preventDefault() in the handleSubmit will prevent the page from refreshing. Instead of executing getRecommendation in the handleSubmit, just redirect to the new page, with the values on state

alexcostaluiz: I'm thinking perhaps it doesn't matter. The only two required fields are latitude and longitude, and in the future these fields will be removed, and the user will choose their location by city/address (or let the browser detect their current location) before they can open the `PreferenceForm`. Once a user gets to the preference form I don't think we will be making any form fields required just in case the user wants to see any nearby restaurants (which would kind of defeat the purpose of the algorithm, but I don't think there's a point in taking away that functionality).

jesseelizabeth: I still think we should have a form onSubmit, just to avoid any undesirable side effects in the future. Having it this way with just a re-route link has the potential to contribute to bugs

alexcostaluiz: Ah okay got it, just seeing your previous message. Will refactor. Thanks!

project/frontend/src/components/Modal.js
```diff
@@ -5,7 +5,7 @@ import React from 'react';
 /**
  * A modal window with a fixed position on the viewport.
  *
- * @param {bool} props.open True if the modal should be shown; false if it
+ * @param {boolean} props.open True if the modal should be shown; false if it
```
kpancha, line 8: nit but should this be `boolean=` like in likes 20 and 22?

project/frontend/src/pages/HomePage.js
```diff
@@ -1,21 +1,27 @@
...
+import Modal from '../components/Modal.js';
+import PreferenceForm from '../components/PreferenceForm.js';
 
 function HomePage() {
-  return (
-    <div className='container u-full-width'>
+  const [isPreferenceFormOpen, setIsPreferenceFormOpen] = useState(false);
+
+  const togglePreferenceForm = () => {
+    setIsPreferenceFormOpen((prev) => !prev);
```
miahsanchez, line 12: how come you can't you just do setIsPreferenceFormOpen(!prev)?

project/frontend/src/components/PreferenceForm.js
kpancha: Adding onto what Jesse said, I would say it would be best to send the entire state object to the MatchResultsPage instead of parsing through each field in the state. Then you can just pass the whole object in as the preferences parameter in getRecommendation

project/frontend/src/pages/MatchResultsPage.js
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
+
+function MatchResultsPage(props) {
+  // Form state is in `props.location.state`.
+
+  return <p>Results page content goes here.</p>;
+}
+
+export default MatchResultsPage;
```
miahsanchez, line 9: what would happen if you changed this back to MatchResults.js ?

project/frontend/src/pages/HomePage.js
alexcostaluiz: setState functions (from a `useState` hook or `setState` in a class component) are asynchronous. So if a `setState` call relies on the previous state it is better to pass a function to `setState` than to pass an object.

This page from Reactjs.org explains it better: https://reactjs.org/docs/faq-state.html. That page is about class components but from what I've read the same thing applies to using the `useState` hook like I've done in this component.

project/frontend/src/components/Modal.js
alexcostaluiz: The `=` indicates that the param has a default value. I could give the `open` prop a default and add a `=` but semantically I think it makes more sense if open doesn't have a default, so that it's clear the parent component should be telling the modal whether it's open or not explicitly. The `center` and `centerHorizontal` props are by default set to false if they are not set explicitly by the parent component.

project/frontend/src/pages/MatchResultsPage.js
alexcostaluiz: Not sure what you mean. I just thing the name `MatchResultsPage` is better semantically because we have `HomePage` and will presumable have more pages like `UserSettingPage` or `SwipeMatchPage` and whatnot.

project/frontend/src/index.js
```diff
@@ -21,10 +21,9 @@ ReactDOM.render(
       <Router>
         <Header />
         <Switch>
-          <Route exact path='/' component={HomePage} />
-          <Route path='/find-match' component={PreferenceForm} />
```
kpancha, line 25: Why did we get rid of the preference form route?

alexcostaluiz: Didn't need it anymore since `PreferenceForm` is now a popup over the `HomePage` component.

project/frontend/src/pages/MatchResultsPage.js
miahsanchez: gotchaaa. I was just wondering if you could change it back to its original name since we're most likely going to have some conflicts and didn't know we had that naming convention!

PR #32, opened 2020-07-10
Implement MatchResultsPage component

alexcostaluiz:
Ready for review!

Here I implemented the MatchResultsPage as it was seen in the demo. I also made a few updates to the RestaurantCard and ImageSlider components.


project/frontend/src/components/RestaurantCard.js
```diff
@@ -10,79 +18,205 @@ import Globe from '../assets/globe.svg';
...
+  return i;
+}
+
+/**
+ * A list of nondescript tags that do not need to be displayed
+ * in the RestaurantCard component.
+ *
+ * @type {!Array<string>}
+ */
+const bannedTags = ['food', 'point_of_interest', 'establishment'];
```
kpancha, line 45: Why are these banned?

project/frontend/src/components/RestaurantCard.js
```diff
@@ -10,79 +18,205 @@ import Globe from '../assets/globe.svg';
...
+
+/**
+ * Modifies the list of tags returned from the Places API into
+ * a more readable format.
+ *
+ * @param {!Array<string>} tags The list of tags for a restaurant.
+ * @return {!Array<string>} An array of more readable tags.
+ */
+function prettifyTags(tags) {
+  const out = [];
```
kpancha, line 55: nit: let's make this var name more descriptive, like `newTagNames` or something

project/frontend/src/components/RestaurantCard.js
alexcostaluiz: I figured they were redundant to show on the restaurant card, since people already know they are looking at a restaurant. `Point of Interest` and `Establishment` are also just super generic so I don't think it adds anything. We can talk about changing this though if you disagree.

PR #33, opened 2020-07-13
Restaurant markers

miahsanchez:
Add the first 4 restaurants as markers on the map! The map is centered around the #1 match.
Screen Shot 2020-07-13 at 2 55 32 PM


project/frontend/src/components/ResultsMap.js
```diff
@@ -10,8 +10,8 @@ const mapStyle = {
 function MapContainer(props) {
   const [activeMarker, changeMarker] = useState({});
   const [showInfoWindow, changeInfoVisibility] = useState(false);
-
-  const coords = { lat: 40.837, lng: -73.865 };
+  const restaurant = props.restaurant;
+  const matchCoords = restaurant[0].key.latLngCoords;
```
kpancha, line 14: Even though we check for an empty list in getRecommendation, I think it might be good to do a check here just to be safe so that we don't get an index out of bounds. Jesse suggested that we make a component called NoResults or something, so we could just redirect to that if restaurat.length === 0

kpancha: I guess we can add that as a todo and right now you could just alert the user and return or something?

project/frontend/src/index.js
```diff
@@ -24,7 +23,6 @@ ReactDOM.render(
           <Route exact path='/' component={HomePage} />
           <Route path='/find-match' component={PreferenceForm} />
           <Route path='/match-results' component={MatchResults} />
-          <Route path='/mapRoute' component={MapContainer} />
```
kpancha, line 27: Make this hyphenated like the other paths rather than camel case I think

miahsanchez: this line was removed in this PR so I don't think there's a point in changing it now!

kpancha: oh oops you're right

project/frontend/src/components/ResultsMap.js
```diff
@@ -10,8 +10,8 @@ const mapStyle = {
 function MapContainer(props) {
   const [activeMarker, changeMarker] = useState({});
   const [showInfoWindow, changeInfoVisibility] = useState(false);
```
alexcostaluiz, line 12: nit: I think the more conventional names for these setter functions would be `setActiveMarker` and `setShowInfoWindow`. I think using the word `set` instead of `change` makes it more clear that these functions are setter functions that take in a parameter for the new value. I think this is mostly a matter of personal preference though, so it's up to you.

miahsanchez: ill change it to the convention!

project/frontend/src/components/ResultsMap.js
miahsanchez: going to add this as a TODO as discussed in our sync today!

project/frontend/src/components/ResultsMap.js
```diff
@@ -8,34 +8,47 @@ const mapStyle = {
 };
 
 function MapContainer(props) {
-  const [activeMarker, changeMarker] = useState({});
-  const [showInfoWindow, changeInfoVisibility] = useState(false);
-
-  const coords = { lat: 40.837, lng: -73.865 };
+  const [activeMarker, setActiveMarker] = useState({});
+  const [showInfoWindow, setShowInfoWindow] = useState(false);
+  // TODO: add a conditional to check if the list is empty
```
kpancha, line 13: Sorry to be picky but I think we should add a basic check in this PR (rather than doing it in a separate PR) just so that nothing breaks-- just something simple like 
`if (!restaurants) { 
    alert('No restaurants found.'); 
    return; 
}` 
and we can come back in a separate PR to actually deal with what happens in this case :)

project/frontend/src/components/ResultsMap.js
```diff
@@ -8,34 +8,47 @@ const mapStyle = {
...
     if (showInfoWindow) {
-      changeInfoVisibility(false);
+      setShowInfoWindow(false);
+    }
+  };
+
+  const createMarkers = () => {
+    let markers = [];
+    const numOfMarkers = 4;
+    for (let i = 0; i < numOfMarkers; i++) {
```
surbhimadan1995, line 31: I think we should change numOfMarkers to min (restaurant.length, 4) -- as restaurant[i] will be not a valid index if we happen to have < 4 restaurants

project/frontend/src/components/ResultsMap.js
miahsanchez: yeah im going to do it in this PR i just haven't done it yet because I was working on another PR which is why i added the TODO

PR #35, opened 2020-07-14
Add location finder to modal using Geocoder

kpancha:
Using Geocoder API for user to enter their current location.

Current functionality:
- text box shows up when user clicks on Location button (find match buttons are disabled until a location has been entered)
- when user clicks the submit "Find Location" button, if the location is found it is pushed to props.history
- if location isn't found it alerts the user
- When user submits their location, text shows the location that was found (example: if I enter "montgomery new jersey" and submit, we display "Montgomery, NJ, USA" back to the user so they can confirm it is the correct location)
- button that can access user's current location without having to enter anything

### DIFF
--- a/project/frontend/src/components/LocationFinder.js
+++ b/project/frontend/src/components/LocationFinder.js
@@ -1,0 +1,59 @@
+import React from 'react';
+
+class LocationFinder extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      locationString: '',
+    };
+    this.changeState = this.changeState.bind(this);
+    this.handleSubmit = this.handleSubmit.bind(this);
+  }
+
+  changeState(event) {
+    this.setState({ locationString: event.target.value });
+  }
+
+  handleSubmit(event) {
+    event.preventDefault();
+    this.getLocation(this.state.locationString);
+  }
+
+  getLocation(userInput) {
+    const proxyUrl = 'https://cors-anywhere.herokuapp.com/';
+    const geocodeBaseUrl = 'https://maps.googleapis.com/maps/api/geocode/json?';
+    const searchParams = new URLSearchParams();
+    searchParams.append('address', userInput);
+    searchParams.append('key', process.env.REACT_APP_GOOGLE_API_KEY);
+    fetch(proxyUrl + geocodeBaseUrl + searchParams, {
+      headers: {
+        'Access-Control-Allow-Origin': '*',
+        'Content-Type': 'application/json',
+      },
+    })
+      .then((response) => response.json())
+      .then((location) => {
+        if (!location) {
+          alert('Could not find any results for that location.');
+          return;
+        }
+        const currLocation = location.results[0].geometry.location;
+        const locationName = location.results[0].formatted_address;
+        this.props.sendData({ currLocation, locationName, });
+      });
+  }
+
+  render() {
+    return(
+      <form onSubmit={this.handleSubmit}>
+        <input
+          name='location'
+          onChange={this.changeState}
+        />
+        <button onClick={this.handleSubmit}>Find Location</button>
+      </form>
+    );
+  }
+  
+}
+export default LocationFinder;
--- a/project/frontend/src/components/LocationFinder.js
+++ b/project/frontend/src/components/LocationFinder.js
@@ -18,7 +18,10 @@ class LocationFinder extends React.Component {
 
   handleSubmit(event) {
     event.preventDefault();
-    this.getLocation(this.state.userInput);
+    const locationData = this.getLocation(this.state.userInput);
+    if (locationData) {
+      this.props.sendData(locationData);
+    }
   }
 
   getLocation(userInput) {
@@ -35,15 +38,16 @@ class LocationFinder extends React.Component {
     })
       .then((response) => response.json())
       .then((location) => {
-        if (!location) {
+        if (location.error_message) {
           alert('Could not find any results for that location.');
-          return;
+          return false;
         }
         const currLocation = location.results[0].geometry.location;
         const locationName = location.results[0].formatted_address;
         this.setState({ locationName });
         this.setState({ submitted: true });
-        this.props.sendData({ currLocation, locationName });
+        
+        return { currLocation, locationName };
       });
   }
 
--- a/project/frontend/src/components/LocationFinder.js
+++ b/project/frontend/src/components/LocationFinder.js
@@ -33,8 +33,7 @@ class LocationFinder extends React.Component {
         geocoder.geocode({ location: currLocation }, (results, status) => {
           if (status === 'OK') {
             const locationName = results[0].formatted_address;
-            this.setState({ locationName });
-            this.setState({ submitted: true });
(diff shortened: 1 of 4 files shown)


project/frontend/src/components/LocationFinder.js
```diff
@@ -0,0 +1,59 @@
...
+  changeState(event) {
+    this.setState({ locationString: event.target.value });
+  }
+
+  handleSubmit(event) {
+    event.preventDefault();
+    this.getLocation(this.state.locationString);
+  }
+
+  getLocation(userInput) {
```
kpancha, line 22: Since some vars (proxyUrl, headers) are common between getLocation here and getRecommendation, I am thinking of moving those functions to the same file and having file level vars for these that can be used

project/frontend/src/components/PreferenceForm.js
```diff
@@ -97,6 +103,8 @@ class PreferenceForm extends React.Component {
     const prices = { Low: 1, Medium: 2, High: 3, 'Very High': 4 };
     return (
       <form onSubmit={this.handleSubmit}>
+        <label htmlFor='location-finder'>Search for your location here</label>
+        <LocationFinder sendData={this.getLocationData} />
```
miahsanchez, line 107: following the format you have right now should this be a child of the label ?

project/frontend/src/components/PreferenceForm.js
```diff
@@ -56,6 +58,10 @@ class PreferenceForm extends React.Component {
     this.setState({ [attrName]: field });
   }
 
+  getLocationData(data) {
```
miahsanchez, line 61: does this return something? why is this a getter function? It seems like it's acting like a setter

kpancha: yep not really a clear name since its kind of doing both but its removed in the latest commit

kpancha: It's getting the data from the LocationFinder but also setting the state, if that makes sense... I agree it is more of a setter though

project/frontend/src/components/LocationFinder.js
alexcostaluiz: I think you should use the Geocoder JavaScript API instead of this herokuapp proxy to get around the CORS errors. You can read about it here https://developers.google.com/maps/documentation/javascript/geocoding. It seems really simple to use.

project/frontend/src/pages/HomePage.js
```diff
@@ -4,23 +4,44 @@ import React, { useState } from 'react';
...
+  const handleClick = () => {
+    togglePreferenceForm();
+    toggleLocationForm();
+  };
+
+  const handleData = (data) => {
+    if (data) {
+      props.history.push(data);
+    }
+  };
```
alexcostaluiz, line 30: Instead of passing data from LocationForm -> HomePage through the `props.history` which is kind of an anti-pattern since props are generally supposed to be immutable but `props.history` is a rare exception to that rule, I think you should declare a another state variable on the HomePage component:

```
const [location, setLocation] = useState({});
```

And then simply pass the `setLocation` function to the `LocationForm` component in a prop so that `LocationForm` can directly update `HomePage`'s state.

project/frontend/src/pages/HomePage.js
```diff
@@ -4,23 +4,44 @@ import React, { useState } from 'react';
...
       <div className='row'>
         <div id='welcome' className='column'>
           <h1 id='logo'>[LOGO]</h1>
           <h4>Discover your restaurant match</h4>
-          <button>Location</button>
+          <button onClick={toggleLocationForm}>Location</button>
           <div>
-            <button onClick={togglePreferenceForm}>Find My Match</button>
+            <button
+              disabled={!Boolean(props.history.location.currLocation)}
```
alexcostaluiz, line 41: So this could instead just be `disabled={!Boolean(location.currLocation)}`.

Also is the `Boolean()` constructor necessary here? Or is it fine to just do `!location.currLocation` and have the `!` make the expression resolve to a boolean?

project/frontend/src/pages/HomePage.js
```diff
@@ -84,6 +105,20 @@ function HomePage(props) {
...
       <PreferenceForm history={props.history} />
     </Modal>,
+    <Modal
+      key='location-form'
+      open={isLocationFormOpen}
+      onDismiss={toggleLocationForm}
+      centerHorizontal={true}
+      top='64px'
+      bottom='64px'>
+      <LocationFinder sendData={handleData} />
```
alexcostaluiz, line 115: And here instead of passing `handleData` we can pass `setLocation` like:
```
<LocationFinder setLocation={setLocation} />
```

project/frontend/src/pages/HomePage.js
kpancha: That makes sense and I think it's needed but can check

project/frontend/src/components/LocationFinder.js
```diff
@@ -0,0 +1,84 @@
...
+            this.setState({ locationName });
+            this.setState({ submitted: true });
+            this.props.sendData({ currLocation, locationName });
+          } else {
+            alert('Geocode was not successful for the following reason: ' + status);
+          }
+        });
+      });
+    } else {
+      alert('Cannot find your location');
```
miahsanchez, line 42: i'm not even sure if we're keeping this alert or if it's a filler right now but if we are could we tell them to try typing in their location?

project/frontend/src/pages/HomePage.js
```diff
@@ -84,6 +103,20 @@ function HomePage(props) {
       bottom='64px'>
       <PreferenceForm history={props.history} />
     </Modal>,
+    <Modal
+      key='location-form'
+      open={isLocationFormOpen}
+      onDismiss={toggleLocationForm}
+      centerHorizontal={true}
+      top='64px'
```
miahsanchez, line 111: I asked this in our chat but is having a set px dynamic for different screen sizes?

kpancha: I don't think it is-- agreed that making it dynamic would better. I just did this to make it the same as the other modal but we can discuss it in our meeting

project/frontend/src/components/LocationFinder.js
kpancha: Good idea! I think we should eventually remove the alerts altogether but for now it would still be helpful to have an alert with more useful information than what I have here lol

project/frontend/src/components/LocationFinder.js
```diff
@@ -0,0 +1,95 @@
...
+      navigator.geolocation.getCurrentPosition((position) => {
+        const currLocation = {
+          lat: position.coords.latitude,
+          lng: position.coords.longitude,
+        };
+        geocoder.geocode({ location: currLocation }, (results, status) => {
+          if (status === 'OK') {
+            const locationName = results[0].formatted_address;
+            this.setState({ locationName });
+            this.setState({ submitted: true });
```
surbhimadan1995, line 37: nit: make this one setState call here as well

PR #37, opened 2020-07-14
Rank appears when you hover over the marker

miahsanchez:
Infowindow with rank (and the restaurant name soon) appear when you scroll over a marker.

I had to copy and paste code from my infoWindows PR that I closed because some really FUNKY git things were happening after I pulled from master but if you wanted to see my old commits check that branch out!

### DIFF
--- a/project/frontend/src/components/ResultsMap.js
+++ b/project/frontend/src/components/ResultsMap.js
@@ -10,13 +10,19 @@ const mapStyle = {
 function MapContainer(props) {
   const [activeMarker, setActiveMarker] = useState({});
   const [showInfoWindow, setShowInfoWindow] = useState(false);
+  const [infoWindowNumber, setInfoWindowNumber] = useState('');
   // TODO: add a conditional to check if the list is empty
   const restaurant = props.restaurant;
   const matchCoords = restaurant[0].key.latLngCoords;
 
-  const onMarkerClick = (props, marker) => {
+  const onMouseOverMarker = (props, marker) => {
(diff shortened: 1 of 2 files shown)


project/frontend/src/index.js
```diff
@@ -25,9 +27,13 @@ ReactDOM.render(
...
         <Switch>
-          <Route exact path='/' component={HomePage} />
-          <Route path='/find-match' component={PreferenceForm} />
-          <Route path='/match-results' component={MatchResults} />
+          <Route path='/card'>
+            <RestaurantCard restaurant={restaurant} details={details} />
+          </Route>
+
+          <Route path='/match-results' component={MatchResultsPage} />
+          <Route path='/mapRoute' component={MapContainer} />
```
kpancha, line 35: Rename to '/map-route'?

project/frontend/src/components/ResultsMap.js
```diff
@@ -10,13 +10,19 @@ const mapStyle = {
...
   const matchCoords = restaurant[0].key.latLngCoords;
 
-  const onMarkerClick = (props, marker) => {
+  const onMouseOverMarker = (props, marker) => {
     setActiveMarker(marker);
     setShowInfoWindow(true);
+    setInfoWindowNumber(marker.id);
+  };
+
+  const onMouseOutMarker = (props) => {
```
kpancha, line 24: I don't think you need props as the param here if its not being used (not sure though and I don't think it's a problem to keep it anyway)

PR #39, opened 2020-07-16
Redo markers

miahsanchez:
The map's center is going to be at the user's location. There is now a conditional that only adds markers once the restaurant list actually has something in it. If the list is larger than size 4 then the first 4 restaurants will have markers!

I moved the restaurant marker code here because there were too many git issues that would not have been resolved in time for prototype demo.

### DIFF
--- a/project/frontend/src/components/ResultsMap.js
+++ b/project/frontend/src/components/ResultsMap.js
@@ -11,34 +11,37 @@ function MapContainer(props) {
   const [activeMarker, setActiveMarker] = useState({});
   const [showInfoWindow, setShowInfoWindow] = useState(false);
   // TODO: add a conditional to check if the list is empty
-  const restaurant = props.restaurant;
-  const matchCoords = restaurant[0].key.latLngCoords;
+  const restaurants = props.restaurants;
+  const userCenter = props.userLocation;
 
   const onMarkerClick = (props, marker) => {
     setActiveMarker(marker);
     setShowInfoWindow(true);
   };
 
-  const onClose = (props) => {
+  const onClose = () => {
     if (showInfoWindow) {
       setShowInfoWindow(false);
     }
   };
 
   const createMarkers = () => {
-    let markers = [];
-    const numOfMarkers = 4;
-    for (let i = 0; i < numOfMarkers; i++) {
-      const coords = restaurant[i].key.latLngCoords;
-      markers.push(
-        <Marker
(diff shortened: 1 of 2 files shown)


project/frontend/src/components/ResultsMap.js
```diff
@@ -3,39 +3,54 @@ import React, { useState } from 'react';
...
-  const [activeMarker, changeMarker] = useState({});
-  const [showInfoWindow, changeInfoVisibility] = useState(false);
-
-  const coords = { lat: 40.837, lng: -73.865 };
+  const [activeMarker, setActiveMarker] = useState({});
+  const [showInfoWindow, setShowInfoWindow] = useState(false);
+  const restaurants = props.restaurants;
+  const userCenter = props.userLocation;
 
   const onMarkerClick = (props, marker) => {
```
surbhimadan1995, line 16: props seems unused - can we remove it?

miahsanchez: for some reason when I remove props here i get an error!

miahsanchez: Screen Shot 2020-07-16 at 1 41 30 PM
^if i remove props

surbhimadan1995: hmm kk, seems like its trying to use something out of props to handle errors internally (just skimming through the doc linked in the error) but thanks for looking!

project/frontend/src/components/ResultsMap.js
```diff
@@ -3,39 +3,54 @@ import React, { useState } from 'react';
...
+          <Marker
+            onClick={onMarkerClick}
+            position={coords}
+            name={'Your #' + toString(i + 1) + ' Match'}
+            aria-label={'Your #' + toString(i + 1) + ' Match'}
+          />
+        );
+      }
+      return markers;
     }
```
surbhimadan1995, line 43: nit: prefer reducing nesting by returning early when possible 
here you can do: 
```
if (restaurants.length == 0) {
    return;
}
let markers = []'
/..... rest of the stuff 
```

kpancha: To be even more concise, rather than `if (restaurants.length == 0) {` you can just use truthy/falsy and say `if (!restaurants) {`

miahsanchez:
closing this PR because im changing to google-map-react so a lot of this code needs to be changed!

PR #40, opened 2020-07-16
added markerHover code

miahsanchez:
Had to move code from markerHover due to all the git issues. When you hover over a marker you now see its ranking and it goes away when you scroll out!

### DIFF
--- a/project/frontend/src/components/ResultsMap.js
+++ b/project/frontend/src/components/ResultsMap.js
@@ -10,12 +10,18 @@ const mapStyle = {
 function MapContainer(props) {
   const [activeMarker, setActiveMarker] = useState({});
   const [showInfoWindow, setShowInfoWindow] = useState(false);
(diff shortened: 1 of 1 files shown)


project/frontend/src/components/ResultsMap.js
```diff
@@ -30,12 +39,16 @@ function MapContainer(props) {
       const numOfMarkers = Math.min(restaurants.length, 4);
       for (let i = 0; i < numOfMarkers; i++) {
         const coords = restaurants[i].key.latLngCoords;
+        const numInList = (i + 1).toString();
```
kpancha, line 42: Do we need to cast this to a string?

miahsanchez: I use this value in strings so i'd rather just call toString once as opposed to every time I reference the variable!

PR #41, opened 2020-07-16
Change react map library from google-maps-react to google-map-react

miahsanchez:
I changed from google-maps-react to google-maps react to avoid the multiple maps error that we have been getting. I now had to create a new marker component and I chose a cute little icon for it :) I still need to add the customized info window but I think I will do this in another PR!


miahsanchez:
Screen Shot 2020-07-16 at 5 58 19 PM
map w markers :')

project/frontend/package.json
```diff
@@ -6,13 +6,15 @@
     "@testing-library/jest-dom": "^4.2.4",
     "@testing-library/react": "^9.3.2",
     "@testing-library/user-event": "^7.1.2",
+    "google-map-react": "^1.1.7",
     "google-maps-react": "^2.0.6",
```
surbhimadan1995, line 10: do we still need this?

alexcostaluiz: +1

PR #43, opened 2020-07-17
infoWindow Component Creation

miahsanchez:
So this is what I have so far for my info window! I followed https://github.com/google-map-react/google-map-react-examples/blob/master/src/examples/MarkerInfoWindow.js for the infoWindow component and haven't really done anything to change it yet besides hard coding in the restaurant data. This is what the card looks like now but I want to discuss with Kira and Alex what we want to be included in the card on Monday!  
Screen Shot 2020-07-17 at 5 56 33 PM

### DIFF
--- a/project/frontend/src/components/ResultsMap.js
+++ b/project/frontend/src/components/ResultsMap.js
@@ -5,9 +5,34 @@ import Lunch from '../assets/lunch.svg';
 function MapContainer(props) {
   const restaurants = props.restaurants;
   const userCenter = props.userLocation;
-  const apple = 5;
 
   const MarkerIcon = () => <img src={Lunch} alt='Lunch icon' />;
+
+  const InfoWindow = (props) => {
+    // TODO: Remove hard coded data and add name and percent match from the backend.
+    const restaurantName = 'Applebees';
+    const percentMatch = '96%';
+    const infoWindowStyle = {
+      position: 'relative',
+      bottom: 80,
+      left: '-45px',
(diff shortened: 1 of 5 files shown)


project/frontend/src/components/ResultsMap.js
```diff
@@ -8,6 +8,46 @@ function MapContainer(props) {
...
+      zIndex: 100,
+    };
+
+    return (
+      <div style={infoWindowStyle}>
+        <div style={{ fontSize: 16 }}>{'Applebees'}</div>
+        <div style={{ fontSize: 14 }}>
+          <span style={{ color: 'grey' }}>{2.5} </span>
+          <span style={{ color: 'orange' }}>
+            {String.fromCharCode(9733).repeat(Math.floor(2.5))}
```
kpancha, line 36: Just curious what this and line 39 are doing (something with the stars?)

miahsanchez: if you look on the card the String.fromCharCode(9733) produces the star so we want the floored rating to have orange stars and the rest of them to be gray. I'm about to push code with var names so certain parts of the card become more clear!

miahsanchez:
Update: the info window now only shows the restaurant name and the percent match
Screen Shot 2020-07-20 at 11 38 58 AM

miahsanchez:
i force pushed after merging master and all of my commits deleted, luckily my code was saved somewhere else

PR #44, opened 2020-07-20
Add material UI multiselect dropdown

kpancha:
Using Zomato API to get cuisines in the area using lat lng (https://developers.zomato.com/documentation#!/common/cuisines). Since API only returns cuisines in the given town, neglecting other cuisines that may be within the chosen radius, so we also allow the user to enter custom cuisines that are not in the autofill.


project/frontend/src/components/PreferenceForm.js
```diff
@@ -84,7 +87,15 @@ class PreferenceForm extends React.Component {
   }
 
   render() {
-    const cuisines = ['Italian ', 'Mexican ', 'Indian '];
+    const classes = makeStyles((theme) => ({
+      root: {
+        width: 500,
+        '& > * + *': {
```
surbhimadan1995, line 93: what does this represent?

project/frontend/yarn.lock
```diff
@@ -1321,12 +1324,25 @@
     react-is "^16.8.0"
     react-transition-group "^4.4.0"
 
+<<<<<<< HEAD
```
surbhimadan1995, line 1327: merge conflict markers ?

project/frontend/src/components/PreferenceForm.js
kpancha: It is for styling (I was following an example from Material UI https://material-ui.com/components/autocomplete/#multiple-values ) but removed this part because it didn't look likw it was doing anything

project/frontend/src/components/PreferenceForm.js
```diff
@@ -72,19 +104,46 @@ class PreferenceForm extends React.Component {
...
+          multiple
+          name='cuisine'
+          id='cuisine'
+          options={this.state.cuisineOptions}
+          onChange={(event, newCuisineList) => {
+            this.setState({ cuisine: newCuisineList });
+          }}
+          filterOptions={(options, params) => {
+            const filtered = filter(options, params);
+            if (params.inputValue !== '' && !options.includes(params.inputValue)) {
```
kpancha, line 127: Thinking of adding something to make this case insensitive (so if I enter "mexican" and "Mexcian" is an entry in options, we won't display a custom option for the user's entry since it is the same word.

project/frontend/src/components/PreferenceForm.js
```diff
@@ -72,19 +106,49 @@ class PreferenceForm extends React.Component {
...
-    console.log(this.state);
     this.props.history.push({
       pathname: '/match-results',
       state: this.state,
     });
   }
 
+  showCuisineOptions() {
+    const filter = createFilterOptions();
+    if (this.state.cuisineOptions) {
```
surbhimadan1995, line 117: optional nit: can reduce some nesting on this return by returning 
```
if (!this.state.cuisineOptions) {
    return null;
}
return ( ...) 
```

project/frontend/src/components/PreferenceForm.js
```diff
@@ -72,19 +106,49 @@ class PreferenceForm extends React.Component {
...
+          )}
+        />
+      );
+    }
+    return null;
+  }
+
   render() {
-    const cuisines = ['Italian ', 'Mexican ', 'Indian '];
+    // const cuisines = ['Italian', 'Mexican', 'Indian'];
```
surbhimadan1995, line 151: ?

project/frontend/src/components/PreferenceForm.js
```diff
@@ -64,27 +108,64 @@ class PreferenceForm extends React.Component {
...
+            if (
+              trimmedInput !== '' &&
+              trimmedInput.length < 25 &&
+              // RegEx to make sure input is only chars and spaces.
+              /^[a-z\s]+$/i.test(trimmedInput) &&
+              // Make sure we don't display duplicates.
+              !(
+                options.includes(trimmedInput) &&
+                options.includes(params.inputValue)
+              )
```
surbhimadan1995, line 149: could we put this in a helper function isInputValid or something that checks all these things and change this to just if (isInputValid())

project/frontend/src/components/PreferenceForm.js
```diff
@@ -24,17 +28,34 @@ class PreferenceForm extends React.Component {
...
+  componentWillUpdate() {
+    const { currLocation } = this.state;
+    if (
+      this.props.currLocation.lat &&
+      this.props.currLocation.lng &&
+      (currLocation['lat'] !== this.props.currLocation.lat ||
+        currLocation['lng'] !== this.props.currLocation.lng)
+    ) {
+      currLocation['lat'] = this.props.currLocation.lat;
+      currLocation['lng'] = this.props.currLocation.lng;
```
jesseelizabeth, line 46: hm this is modifying state directly? might want to pull these out to const newLat = this.props.currLocation.lat

jesseelizabeth: (not actually modifying state directly, but it's slightly confusing - totally optional though)

kpancha: So are you suggesting then setting `currLocation['lat'] = newLat`?

PR #45, opened 2020-07-21
Implement swipe match backend

alexcostaluiz:
draft


project/backend/src/main/java/com/google/sps/servlets/SwipeMatchServlet.java
```diff
@@ -0,0 +1,124 @@
...
+
+        for (AsyncContext asyncContext : asyncContexts.values()) {
+          try {
+            sendUser(asyncContext.getResponse().getWriter(), user);
+          } catch (Exception e) {
+            asyncContexts.values().remove(asyncContext);
+          }
+        }
+      } catch (InterruptedException e) {
+        // idk yet
```
kpancha, line 47: What is an example of when this block would execute (to help us figure out how to handle the exception)?

project/backend/src/main/java/com/google/sps/servlets/SwipeMatchServlet.java
```diff
@@ -0,0 +1,124 @@
...
+    }
+  }
+
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    String username = request.getParameter("username");
+    try {
+      queue.put(username);
+    } catch (InterruptedException e) {
+      // failed, idk yet
```
kpancha, line 114: We could probably log this with Logger, at least for now

project/backend/src/main/java/com/google/sps/servlets/SwipeMatchServlet.java
```diff
@@ -0,0 +1,124 @@
...
+          }
+        }
+      } catch (InterruptedException e) {
+        // idk yet
+      }
+    }
+  });
+
+  @Override
+  public void destroy() {
```
miahsanchez, line 53: when would we be calling this?

alexcostaluiz:
This method of implementing swipe match did not work with our appengine setup.

PR #46, opened 2020-07-21
Implement swipe match frontend

alexcostaluiz:
draft


project/frontend/src/pages/SwipeMatchPage.js
```diff
@@ -0,0 +1,34 @@
+import React, { useState, useEffect } from 'react';
+
+function SwipeMatchPage(props) {
+  const username = props.location.state.username;
```
miahsanchez, line 4: where is the username coming from? the profile page or is this going to be a user input?

alexcostaluiz:
This method of implementing swipe match did not work with our appengine setup.

PR #47, opened 2020-07-21
Hover window

miahsanchez:
When you scroll over a marker an infoWindow pops up (still with the hardcoded data). And disappears when you scroll out.


project/frontend/src/components/ResultsMap.js
```diff
@@ -1,12 +1,55 @@
...
+      position: 'relative',
+      bottom: 80,
+      left: '-45px',
+      width: 220,
+      backgroundColor: 'white',
+      boxShadow: '0 2px 7px 1px rgba(0, 0, 0, 0.3)',
+      padding: 10,
+      fontSize: 14,
+      zIndex: 100,
+    };
```
miahsanchez, line 42: for some reason i put that im comparing this branch to my addInfoWindows branch which has this code but it still says that this is new to this branch ? confusion

jesseelizabeth: Did you branch from the addInfoWindows branch to create this branch?

miahsanchez: I did!

project/frontend/src/components/ResultsMap.js
```diff
@@ -1,12 +1,54 @@
...
 
 function MapContainer(props) {
   const restaurants = props.restaurants;
   const userCenter = props.userLocation;
+  const [activeMarker, setActiveMarker] = useState({});
+  const [showInfoWindows, setShowInfoWindows] = useState([
+    false,
+    false,
+    false,
+    false,
```
surbhimadan1995, line 13: for this, I think we should add the param name for the state to make it clear what the state is referring to 
name1: false,
name2: false 

etc

PR #51, opened 2020-07-23
Hover Window Copy

miahsanchez:
Copied my code from the `hoverWindow` branch since git was making it impossible to do anything.


project/frontend/src/components/ResultsMap.js
```diff
@@ -1,12 +1,28 @@
...
 function MapContainer(props) {
   const restaurants = props.restaurants;
   const userCenter = props.userLocation;
+  const [activeMarker, setActiveMarker] = useState({});
+  const [showInfoWindows, setShowInfoWindows] = useState([
+    false,
+    false,
+    false,
+    false,
+  ]);
```
surbhimadan1995, line 14: for this, I think we should add the param name for the state to make it clear what the state is referring to
name1: false,
name2: false

etc

PR #52, opened 2020-07-23
Add Name and Percent Copy

miahsanchez:
I copied the code from the `addNameandPercent` branch which gets the name and match percent of the restaurant from the backend and displays that on the infoWindow

### DIFF
--- a/project/frontend/src/components/ResultsMap.js
+++ b/project/frontend/src/components/ResultsMap.js
@@ -1,17 +1,41 @@
-import React from 'react';
+import React, { Fragment, useState } from 'react';
 import GoogleMapReact from 'google-map-react';
 import Lunch from '../assets/lunch.svg';
 
 function MapContainer(props) {
   const restaurants = props.restaurants;
   const userCenter = props.userLocation;
+  const [activeMarker, setActiveMarker] = useState({});
+  const [showInfoWindows, setShowInfoWindows] = useState({
+    marker0: false,
+    marker1: false,
+    marker2: false,
+    marker3: false,
+  });
 
-  const MarkerIcon = () => <img src={Lunch} alt='Lunch icon' />;
+  // TODO: Add marker for "you are here location".
+  const MarkerIcon = (props) => {
+    const markerIndex = props.id;
+    const currentRestaurant = restaurants[markerIndex];
+    const percentValue = Math.round(currentRestaurant.value * 100);
+    return (
+      <Fragment>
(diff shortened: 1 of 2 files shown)


project/frontend/src/components/ResultsMap.js
```diff
@@ -7,27 +7,36 @@ function MapContainer(props) {
...
       </Fragment>
     );
   };
 
   const InfoWindow = (props) => {
-    // TODO: Remove hard coded data and add name and percent match from the backend.
-    const restaurantName = 'Applebees';
-    const percentMatch = '96%';
+    const restaurantName = props.restaurantName;
+    const percentMatch = props.percentMatch;
```
jesseelizabeth, line 39: nit: can change this to `const {restaurantName, percentMatch} = props`

project/frontend/src/components/ResultsMap.js
```diff
@@ -7,27 +7,36 @@ function MapContainer(props) {
...
     marker2: false,
     marker3: false,
-    marker4: false,
   });
 
   // TODO: Add marker for "you are here location".
   const MarkerIcon = (props) => {
-    const markerName = 'marker' + props.id;
+    const markerID = props.id;
+    const markerName = 'marker' + markerID;
```
jesseelizabeth, line 19: nit: markerID seems pretty self explanatory - you could maybe just simplify this to `const makerName = `marker ${props.id}``

jesseelizabeth: the tick marks there need to be back ticks `

jesseelizabeth: But it's a handy trick when creating chained strings with variables so you do have to manually add + and extra spaces etc
read more about it here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

miahsanchez: this is good to know!! thank you :)

project/frontend/src/components/ResultsMap.js
```diff
@@ -15,19 +15,27 @@ function MapContainer(props) {
 
   // TODO: Add marker for "you are here location".
   const MarkerIcon = (props) => {
-    const markerName = 'marker' + props.id;
+    const markerID = props.id;
+    const markerName = `marker${props.id}`;
```
jesseelizabeth, line 19: Didn't realize you were also using variable `markerID` below. Since you are you can use it here as well instead of props.id

PR #53, opened 2020-07-23
Render markers based on the current marker index!

miahsanchez:
Changed the createMarkers method so that the markers on the screen correspond to the current restaurant card, the three markers before that in the restaurant matches list and the three markers after that. And the currentCardIndex changes when a marker is clicked on which will switch the current restaurant card on display and re-render the markers :)

### DIFF
--- a/project/frontend/src/components/ResultsMap.js
+++ b/project/frontend/src/components/ResultsMap.js
@@ -12,9 +12,9 @@ function MapContainer(props) {
     marker1: false,
     marker2: false,
     marker3: false,
-    // marker4: false,
-    // marker5: false,
-    // marker6: false,
+    marker4: false,
+    marker5: false,
+    marker6: false,
   });
 
   // TODO: Add marker for "you are here location".
--- a/project/frontend/src/components/ResultsMap.js
+++ b/project/frontend/src/components/ResultsMap.js
@@ -3,9 +3,17 @@ import GoogleMapReact from 'google-map-react';
 import Lunch from '../assets/lunch.svg';
 
 function MapContainer(props) {
-  const restaurants = props.restaurants;
-  const { currentCardIndex, setCurrentCardIndex } = props;
-  const userCenter = props.userLocation;
+  const {
+    restaurants,
+    currentCardIndex,
+    setCurrentCardIndex,
+    userLocation,
+  } = props;
+  const center =
+    restaurants.length === 0
+      ? userLocation
+      : restaurants[currentCardIndex].key.latLngCoords;
+  const [mapCenter, setMapCenter] = useState(center);
   const [activeMarker, setActiveMarker] = useState({});
   const [showInfoWindows, setShowInfoWindows] = useState({
     marker0: false,
@@ -121,13 +129,14 @@ function MapContainer(props) {
 
   const onMouseClickMarker = (props, marker) => {
     setCurrentCardIndex(marker.id);
+    setMapCenter(restaurants[currentCardIndex].key.latLngCoords);
   };
 
   const mapStyle = { height: '100vh', width: '50%' };
   return (
     <GoogleMapReact
       bootstrapURLKeys={{ key: process.env.REACT_APP_GOOGLE_API_KEY }}
-      defaultCenter={userCenter}
+      center={mapCenter}
       defaultZoom={10}
       style={mapStyle}
       onChildMouseEnter={onMouseEnterMarker}
--- a/project/frontend/src/components/ResultsMap.js
+++ b/project/frontend/src/components/ResultsMap.js
@@ -14,7 +14,6 @@ function MapContainer(props) {
       ? userLocation
       : restaurants[currentCardIndex].key.latLngCoords;
   const [mapCenter, setMapCenter] = useState(center);
-  const [activeMarker, setActiveMarker] = useState({});
   const [showInfoWindows, setShowInfoWindows] = useState({
     marker0: false,
     marker1: false,
@@ -37,7 +36,6 @@ function MapContainer(props) {
         <img src={Lunch} alt={'lunch icon'} />
         {showInfoWindows[markerName] && (
           <InfoWindow
-            marker={activeMarker}
             restaurantName={restaurantName}
             percentMatch={percentValue + '%'}
           />
@@ -76,7 +74,7 @@ function MapContainer(props) {
     if (restaurants.length === 0) {
       return;
     }
-    let markers = [];
+    const markers = [];
     let coords = restaurants[currentCardIndex].key.latLngCoords;
     markers.push(
       <MarkerIcon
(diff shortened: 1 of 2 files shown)


project/frontend/src/components/ResultsMap.js
```diff
@@ -59,23 +71,37 @@ function MapContainer(props) {
...
+    for (let i = 1; i <= numOfMarkers; i++) {
+      const afterIndex = currentCardIndex + i;
+      if (afterIndex < restaurants.length) {
+        coords = restaurants[afterIndex].key.latLngCoords;
+        markers.push(
+          <MarkerIcon lat={coords.lat} lng={coords.lng} id={afterIndex} />
+        );
+      }
+      const beforeIndex = currentCardIndex - i;
+      if (!(beforeIndex < 0)) {
```
alexcostaluiz, line 99: nit: change to `>=` and remove `!`

project/frontend/src/components/ResultsMap.js
```diff
@@ -59,23 +71,37 @@ function MapContainer(props) {
...
-          aria-label={'Your #' + numInList + ' Match'}
-        />
-      );
+    let coords = restaurants[currentCardIndex].key.latLngCoords;
+    markers.push(
+      <MarkerIcon
+        lat={coords.lat}
+        lng={coords.lng}
+        id={currentCardIndex}
+        aria-label={'Your #' + currentCardIndex + ' Match'}
```
alexcostaluiz, line 86: This will be "Your #0 Match" if currentCardIndex is 0. Maybe change to `aria-label={'Your #' + (currentCardIndex + 1) + ' Match'}`

project/frontend/src/components/ResultsMap.js
```diff
@@ -59,23 +68,37 @@ function MapContainer(props) {
...
+      />
+    );
+    const numOfMarkers = Math.min(restaurants.length, 3);
+    for (let i = 1; i <= numOfMarkers; i++) {
```
surbhimadan1995, line 87: hmm I don't quite follow why this loop is [1, numMarkers] instead of [0, numMarkers) -- maybe add an explanation if it has to be the former?

project/frontend/src/components/ResultsMap.js
```diff
@@ -59,23 +68,37 @@ function MapContainer(props) {
...
+      const afterIndex = currentCardIndex + i;
```
surbhimadan1995, line 88: naming nit: lets go with nextCardIndex and previousCardIndex instead of after/before

project/frontend/src/components/ResultsMap.js
```diff
@@ -59,23 +68,37 @@ function MapContainer(props) {
...
+      if (afterIndex < restaurants.length) {
```
surbhimadan1995, line 89: for both this check and the before >= 0 check, I think we should abstract away checking if index is valid into a helper function:
```
function isValidIndex(int index):
    return index >= 0 && index < restaurants.length;
```
then you can replace these checks with if isValidIndex(nextCardIndex) & if isValidIndex(prevCardIndex)

project/frontend/src/components/ResultsMap.js
miahsanchez: Right before the for loop I added the current marker (which is the marker corresponding to the restaurant card that the user is on). In the for loop I want to add the 3 markers before the current marker and the 3 markers after so to do this I subtract and add 1, 2, and 3 from the current index and add those restaurant markers which is why I exclude 0. I can put this into a comment now if you think that this logic makes sense!

miahsanchez: sounds good to me! Is there a reason that it's delta instead of i? just curious

project/frontend/src/components/ResultsMap.js
```diff
@@ -1,19 +1,29 @@
...
+      ? userLocation
+      : restaurants[currentCardIndex].key.latLngCoords;
   const [showInfoWindows, setShowInfoWindows] = useState({
     marker0: false,
     marker1: false,
     marker2: false,
     marker3: false,
+    marker4: false,
+    marker5: false,
+    marker6: false,
```
surbhimadan1995, line 24: Is there a way to create this using a loop? If we wanted to change the # of markers or changed the delta of 3 below where we create them, this seems like it'd be prone to falling out of sync

project/frontend/src/components/ResultsMap.js
surbhimadan1995: oh just for naming clarity for what it represents (imo 'i' here would not be as descriptive of the fact that its actually the # of steps before or after the current index and not all loop indices have to be i :D ) 
if you want to come up with a better name though, feel free to go for it -- delta is just sort of in my head a way to think about diff from a certain point

project/frontend/src/components/ResultsMap.js
kpancha: +1 to this! You could do a loop from 0 to numOfMarkers and then just add a key value pair to the object in each iteration

project/frontend/src/components/ResultsMap.js
```diff
@@ -58,34 +67,71 @@ function MapContainer(props) {
...
+      }
     }
     return markers;
   };
 
+  /** Checks that the current index is a valid index for
+   * the restaurant matches list.
+   */
+  const isValidIndex = (index) => {
+    return index >= 0 && index < restaurants.length;
```
kpancha, line 126: Super nit but since this is just a one liner I think we can make it `const isValidIndex = (index) => index >= 0 && index < restaurants.length;`

project/frontend/src/components/ResultsMap.js
miahsanchez: it turns out i don't even need a for loop! The functions that I have for onMouseEnterMarker and onMouseLeaveMarker handle adding these properties! And is dynamic depending on the size of the restaurant list.

project/frontend/src/components/ResultsMap.js
```diff
@@ -58,34 +59,46 @@ function MapContainer(props) {
...
-        <MarkerIcon
-          lat={coords.lat}
-          lng={coords.lng}
-          id={i}
-          aria-label={'Your #' + numInList + ' Match'}
-        />
-      );
+    const markers = [];
+    const numOfMarkers = Math.min(restaurants.length, 3);
+    for (let delta = -3; delta <= numOfMarkers; delta++) {
```
surbhimadan1995, line 74: super nit: so number of markers is technically 7 here (-3 --> 3) so the naming is a bit misleading , wdyt of 
const numAdjacentMarkers = 3; and then just having the loop (delta = -numAdjacentMarkers; delta <= numAdjacentMarkers, delta++) ? 
I think we can at this point get rid of the min(restaurants.length, 3) check as now isValidIndex should handle that and if length < 3 then we end up showing all of them 

oh and also on the header comment (line 64-67) replace "3" with "numAdjacentMarkers" -- comments can fall out of sync with the actual values used so might as well leave it variable:)

miahsanchez: this is great! I was thinking numOfMarkers didn't make sense here anymore thank you!

PR #55, opened 2020-07-24
Add unit tests

kpancha:
Unit tests for JsonToRestaurantParser and RestaurantScorer

### DIFF
--- a/project/backend/src/test/java/com/google/sps/JsonToRestaurantParserTest.java
+++ b/project/backend/src/test/java/com/google/sps/JsonToRestaurantParserTest.java
@@ -10,8 +10,6 @@ import java.util.List;
 import java.util.Map;
 import org.json.JSONException;
 import org.json.JSONObject;
-import org.junit.Assert;
-import org.junit.Before;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.junit.runners.JUnit4;
@@ -52,7 +50,7 @@ public final class JsonToRestaurantParserTest {
       PLACE_ID, RESTAURANT_NAME, VICINITY, COORDS, RATING, NUM_RATINGS,
       PRICE_LEVEL, EMPTY_TYPES);
     Restaurant actual = JsonToRestaurantParser.toRestaurant(body);
-    Assert.assertEquals(actual, expected);
+    assertThat(actual).isEqualTo(expected);
   }
 
   @Test
@@ -89,7 +87,7 @@ public final class JsonToRestaurantParserTest {
       PLACE_ID, RESTAURANT_NAME, VICINITY, COORDS, /* avgRating= */ MISSING_FIELD, 
       /* numRatings= */ MISSING_FIELD, /* priceNum= */ MISSING_FIELD, EMPTY_TYPES);
     Restaurant actual = JsonToRestaurantParser.toRestaurant(body);
-    Assert.assertEquals(actual, expected);
+    assertThat(actual).isEqualTo(expected);
   }
 
   @Test(expected = JSONException.class)
--- a/project/backend/src/test/java/com/google/sps/JsonToRestaurantParserTest.java
+++ b/project/backend/src/test/java/com/google/sps/JsonToRestaurantParserTest.java
@@ -85,8 +85,7 @@ public final class JsonToRestaurantParserTest {
 
   @Test(expected = JSONException.class)
   public void noFields() throws JSONException {
-    JSONObject body = new JSONObject();
-    Restaurant actual = JsonToRestaurantParser.toRestaurant(body);
+    Restaurant actual = JsonToRestaurantParser.toRestaurant(new JSONObject());
   }
 
   @Test(expected = JSONException.class)
@@ -96,6 +95,6 @@ public final class JsonToRestaurantParserTest {
       .put("name", RESTAURANT_NAME)
       .put("vicinity", VICINITY)
       .put("types", EMPTY_TYPES);
-    Restaurant actual = JsonToRestaurantParser.toRestaurant(body);
+    JsonToRestaurantParser.toRestaurant(body);
   }
 }
--- a/project/backend/src/test/java/com/google/sps/JsonToRestaurantParserTest.java
+++ b/project/backend/src/test/java/com/google/sps/JsonToRestaurantParserTest.java
@@ -4,6 +4,7 @@ import com.google.sps.data.JsonToRestaurantParser;
 import com.google.sps.data.Restaurant;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import org.json.JSONException;
@@ -26,10 +27,10 @@ public final class JsonToRestaurantParserTest {
   private static final String RESTAURANT_NAME = "RESTAURANT_A";
   private static final String PLACE_ID = "A12345";
   private static final String VICINITY = "10 Main Street";
-  private static final Map<String, Double> COORDS = Map.ofEntries(
-    Map.entry("lat", 10.0),
-    Map.entry("lng", 20.0)
-  );
+  private static final Map<String, Double> COORDS = new HashMap<String, Double>() {{
+    put("lat", 10.0);
+    put("lng", 20.0);
+  }};
 
   @Test
   public void allRequiredAndOptionalFields() throws JSONException {
--- a/project/backend/src/test/java/com/google/sps/JsonToRestaurantParserTest.java
+++ b/project/backend/src/test/java/com/google/sps/JsonToRestaurantParserTest.java
@@ -44,11 +44,11 @@ public final class JsonToRestaurantParserTest {
       .put("user_ratings_total", NUM_RATINGS)
       .put("price_level", PRICE_LEVEL);
 
-    Restaurant expected = Restaurant.create(
+    Restaurant expectedRestaurant = Restaurant.create(
       PLACE_ID, RESTAURANT_NAME, VICINITY, COORDS, RATING, NUM_RATINGS,
       PRICE_LEVEL, EMPTY_TYPES);
-    Restaurant actual = JsonToRestaurantParser.toRestaurant(body);
-    assertThat(actual).isEqualTo(expected);
+    Restaurant actualRestaurant = JsonToRestaurantParser.toRestaurant(body);
+    assertThat(actualRestaurant).isEqualTo(expectedRestaurant);
   }
 
   @Test
@@ -62,8 +62,8 @@ public final class JsonToRestaurantParserTest {
       .put("types", randomTypes)
       .put("geometry", geometry);
 
-    Restaurant actual = JsonToRestaurantParser.toRestaurant(body);
-    assertThat(actual.getPlaceTypes()).containsExactlyElementsIn(ALL_TYPES);
+    Restaurant actualRestaurant = JsonToRestaurantParser.toRestaurant(body);
+    assertThat(actualRestaurant.getPlaceTypes()).containsExactlyElementsIn(ALL_TYPES);
   }
 
   @Test
@@ -76,16 +76,16 @@ public final class JsonToRestaurantParserTest {
       .put("types", EMPTY_TYPES)
       .put("geometry", geometry);
 
-    Restaurant expected = Restaurant.create(
+    Restaurant expectedRestaurant = Restaurant.create(
       PLACE_ID, RESTAURANT_NAME, VICINITY, COORDS, /* avgRating= */ MISSING_FIELD, 
       /* numRatings= */ MISSING_FIELD, /* priceNum= */ MISSING_FIELD, EMPTY_TYPES);
-    Restaurant actual = JsonToRestaurantParser.toRestaurant(body);
-    assertThat(actual).isEqualTo(expected);
+    Restaurant actualRestaurant = JsonToRestaurantParser.toRestaurant(body);
+    assertThat(actualRestaurant).isEqualTo(expectedRestaurant);
   }
 
   @Test(expected = JSONException.class)
   public void noFields() throws JSONException {
-    Restaurant actual = JsonToRestaurantParser.toRestaurant(new JSONObject());
+    JsonToRestaurantParser.toRestaurant(new JSONObject());
   }
 
   @Test(expected = JSONException.class)
--- a/project/backend/src/test/java/com/google/sps/JsonToRestaurantParserTest.java
+++ b/project/backend/src/test/java/com/google/sps/JsonToRestaurantParserTest.java
@@ -33,8 +33,6 @@ public final class JsonToRestaurantParserTest {
 
   @Test
   public void allRequiredAndOptionalFields() throws JSONException {
-    // JSON object has all required and all optional fields.
-    // We should see a restaurant with all the values in the JSON object.
     JSONObject geometry = new JSONObject().put("location", COORDS);
     JSONObject body = new JSONObject()
       .put("place_id", PLACE_ID)
@@ -54,9 +52,7 @@ public final class JsonToRestaurantParserTest {
   }
 
   @Test
-  public void extraneousTypes() throws JSONException {
-    // JSON object has types that should be disregarded.
-    // We should see a restaurant with only the three relevant types.
+  public void extraneousTypes_onlyIncludeRelevantTypes() throws JSONException {
     List<String> randomTypes = Arrays.asList(new String[]{"type a", "meal_takeaway", "type b", "meal_delivery", "restaurant"});
     JSONObject geometry = new JSONObject().put("location", COORDS);
     JSONObject body = new JSONObject()
@@ -71,10 +67,7 @@ public final class JsonToRestaurantParserTest {
   }
 
   @Test
-  public void noOptionalFields() throws JSONException {
-    // JSON object has all required fields and no optional fields.
-    // We should see a restaurant that has the required fields and 
-    // default values for optional fields.
+  public void noOptionalFields_useDefaultValues() throws JSONException {
     JSONObject geometry = new JSONObject().put("location", COORDS);
     JSONObject body = new JSONObject()
       .put("place_id", PLACE_ID)
@@ -92,16 +85,12 @@ public final class JsonToRestaurantParserTest {
 
   @Test(expected = JSONException.class)
   public void noFields() throws JSONException {
-    // JSON object has no fields.
-    // We should see a JSONException thrown.
     JSONObject body = new JSONObject();
     Restaurant actual = JsonToRestaurantParser.toRestaurant(body);
   }
 
   @Test(expected = JSONException.class)
   public void missingMandatoryField() throws JSONException {
-    // JSON object is missing a required field.
-    // We should see a JSONException thrown.
     JSONObject body = new JSONObject()
       .put("place_id", PLACE_ID)
       .put("name", RESTAURANT_NAME)
--- a/project/backend/src/test/java/com/google/sps/JsonToRestaurantParserTest.java
+++ b/project/backend/src/test/java/com/google/sps/JsonToRestaurantParserTest.java
@@ -3,6 +3,7 @@ package com.google.sps;
 import com.google.sps.data.JsonToRestaurantParser;
 import com.google.sps.data.Restaurant;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 import org.json.JSONException;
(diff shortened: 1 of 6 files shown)


project/backend/src/test/java/com/google/sps/TestRestaurantParser.java
```diff
@@ -0,0 +1,43 @@
...
+
+  @Test
+  public void noOptionalFields() throws JSONException {
+    JSONObject body = new JSONObject(
+      "{'geometry' : {'location' : {'lat' : " + COORDS_A.get("lat") + ", 'lng' : " + COORDS_A.get("lng") + "}}," + 
+      "'place_id' : " + PLACE_ID_A + 
+      ",'name' : " + RESTAURANT_A + 
+      ",'vicinity' : " + VICINITY_A + 
+      ",'types' : [],}"
+    );
```
kpancha, line 37: Not sure if I'm happy with the way I initialized this... thinking using the way they did it in 4.1 here may be better https://www.baeldung.com/java-org-json#1-creating-json-directly-from-jsonobject

project/backend/src/test/java/com/google/sps/TestRestaurantParser.java
```diff
@@ -0,0 +1,43 @@
...
+  private static final String RESTAURANT_A = "RESTAURANT_A";
+  private static final String PLACE_ID_A = "A12345";
+  private static final String VICINITY_A = "10 Main Street";
+  private static final Map<String, Double> COORDS_A = Map.ofEntries(
+    Map.entry("lat", 10.0),
+    Map.entry("lng", 20.0)
+  );
+
+  @Test
+  public void noOptionalFields() throws JSONException {
```
kpancha, line 30: Also I believe this exception should be caught

project/backend/src/test/java/com/google/sps/TestRestaurantParser.java
```diff
@@ -0,0 +1,45 @@
...
+import org.json.JSONException;
+import org.json.JSONObject;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+@RunWith(JUnit4.class)
+public final class TestRestaurantParser {
```
surbhimadan1995, line 17: file naming nit: it should be named ClassNameTest 
so in this case JsonToRestaurantParserTest.java

project/backend/src/test/java/com/google/sps/TestRestaurantParser.java
surbhimadan1995: nah, so for tests, you usually just let this be and the caller will handle the exception, a test that should not throw an exception, will just pass if what you're trying to assert at the end passes

surbhimadan1995: you should have a separate test where you assert that the exception is thrown to test that case (Where you give it a malformed input)

project/backend/src/test/java/com/google/sps/TestRestaurantParser.java
```diff
@@ -0,0 +1,45 @@
...
+  @Test
+  public void noOptionalFields() throws JSONException {
+    JSONObject body = new JSONObject();
+    JSONObject geometry = new JSONObject();
+    geometry.put("location", COORDS_A);
+    body.put("place_id", PLACE_ID_A);
+    body.put("name", RESTAURANT_A);
+    body.put("vicinity", VICINITY_A);
+    body.put("types", EMPTY_TYPES);
+    body.put("geometry", geometry);
```
surbhimadan1995, line 38: so you can chain these put calls because the method returns the object back to you https://www.javadoc.io/doc/org.json/json/20170516/org/json/JSONObject.html#put-java.lang.String-java.lang.Object- (for the purposes of chaining) 

```
JSONObject geometry = new JSONObject().put(location, coordsA);
JSONObject body = new JSONObject().put(...).put(...).put(...); 
```

project/backend/src/test/java/com/google/sps/TestRestaurantParser.java
```diff
@@ -0,0 +1,45 @@
...
+    
+    Restaurant expected = Restaurant.create(
+      PLACE_ID_A, RESTAURANT_A, VICINITY_A, COORDS_A, -1, -1, -1, EMPTY_TYPES);
```
surbhimadan1995, line 41: add param tags to the -1s

project/backend/src/test/java/com/google/sps/JsonToRestaurantParserTest.java
```diff
@@ -0,0 +1,136 @@
...
+    JSONObject body = new JSONObject()
+      .put("place_id", PLACE_ID)
+      .put("name", RESTAURANT_NAME)
+      .put("vicinity", VICINITY)
+      .put("types", randomTypes)
+      .put("geometry", geometry);
+
+    Restaurant expected = Restaurant.create(
+      PLACE_ID, RESTAURANT_NAME, VICINITY, COORDS, /* avgRating= */ -1, 
+      /* numRatings= */ -1, /* priceNum= */ -1, ALL_TYPES);
```
surbhimadan1995, line 72: should we use MISSING_FIELD for all the -1s here?

project/backend/src/test/java/com/google/sps/JsonToRestaurantParserTest.java
```diff
@@ -0,0 +1,136 @@
...
+    Restaurant actual = JsonToRestaurantParser.toRestaurant(body);
+    Assert.assertTrue(actual.getPlaceTypes().size() == ALL_TYPES.size()
```
surbhimadan1995, line 74: consider using truth for fluent assertions: https://truth.dev/ 
this will help make the assertions clearer and readable

kpancha: Are you suggesting that I use this for all assertions or just this one/ more complicated ones?

surbhimadan1995: for list equality type things, def , and I think even for simpler ones since we're using it anyway 
assertThat(actual).isEqualTo(expected); reads nicer than Assert.assertEquals(actual, expected); I think 
Truth is very widely used in most our junit testing because it reads more like an english sentence :D

surbhimadan1995: check out the javadocs , here's the ones for assertions on list/ set/ map elements: https://truth.dev/api/0.25/com/google/common/truth/IterableSubject.html 

so for this example, it can just be:
assertThat(actual.getPlaceTypes()).containsExactlyElementsIn(ALL_TYPES) ;

surbhimadan1995: and if you care about checking that the order matches assertThat(actual.getPlaceTypes()).containsExactlyElementsIn(ALL_TYPES).inOrder():

project/backend/src/test/java/com/google/sps/JsonToRestaurantParserTest.java
```diff
@@ -0,0 +1,112 @@
...
+      PLACE_ID, RESTAURANT_NAME, VICINITY, COORDS, /* avgRating= */ MISSING_FIELD, 
+      /* numRatings= */ MISSING_FIELD, /* priceNum= */ MISSING_FIELD, EMPTY_TYPES);
+    Restaurant actual = JsonToRestaurantParser.toRestaurant(body);
+    assertThat(actual).isEqualTo(expected);
+  }
+
+  @Test(expected = JSONException.class)
+  public void noFields() throws JSONException {
+    // JSON object has no fields.
+    // We should see a JSONException thrown.
```
surbhimadan1995, line 96: nit: I don;t think these comments are needed as the expected= and test method name says exactly this 
(consider comment usefulness by looking at for instance - does it just describe the method name or what the code is doing or do I need a comment to explain why something was done a certain way?)

project/backend/src/test/java/com/google/sps/JsonToRestaurantParserTest.java
```diff
@@ -0,0 +1,112 @@
...
+      .put("vicinity", VICINITY)
+      .put("types", randomTypes)
+      .put("geometry", geometry);
+
+    Restaurant actual = JsonToRestaurantParser.toRestaurant(body);
+    assertThat(actual.getPlaceTypes()).containsExactlyElementsIn(ALL_TYPES);
+  }
+
+  @Test
+  public void noOptionalFields() throws JSONException {
```
surbhimadan1995, line 74: so for these cases consider a naming pattern along the lines of 
situationBeingTested_expectedOutcome 
-- in this specific case, that could translate to 
noOptionalFields_usesDefaultsForOptional 
^ which eliminates the need for "what this test is doing" comments and makes it self-explanatory 

This feels like it can get verbose with test naming and setup and that sort of thing is more ok in unit testing than actual code. On a related note - we often talk about DRY (dont repeat yourself) when we refer to actual code where we encourage helper methods and abstractions when needed to avoid repitition. 
But for tests, often, DRY is not encouraged https://stackoverflow.com/questions/6453235/what-does-damp-not-dry-mean-when-talking-about-unit-tests. -- and instead we choose Descriptive and meaningful phrases 

of course, there's no one size fits all so usually tests end up being on the spectrum of dry --> damp but in general with unit testing, some repetition (such as with setting up inputs in a given test etc) is not just fine but encouraged as we usually want to make tests as self-contained as possible with distinct "setup" "do the work" "assert the outcome" blocks

kpancha: So if I use that naming pattern, should I do it for all the tests or only the ones that it seems necessary for? For example the noFields test seems self explanatory of what is expected to happen so would I want to still follow that naming convention there?

surbhimadan1995: yep - only ones where the expected outcome isn't obvious without comments - noFields seems fine as is!

project/backend/src/test/java/com/google/sps/JsonToRestaurantParserTest.java
```diff
@@ -0,0 +1,101 @@
...
+      PLACE_ID, RESTAURANT_NAME, VICINITY, COORDS, /* avgRating= */ MISSING_FIELD, 
+      /* numRatings= */ MISSING_FIELD, /* priceNum= */ MISSING_FIELD, EMPTY_TYPES);
+    Restaurant actual = JsonToRestaurantParser.toRestaurant(body);
+    assertThat(actual).isEqualTo(expected);
+  }
+
+  @Test(expected = JSONException.class)
+  public void noFields() throws JSONException {
+    JSONObject body = new JSONObject();
+    Restaurant actual = JsonToRestaurantParser.toRestaurant(body);
```
surbhimadan1995, line 89: nit: this can be reduced to just 
```
JsonToRestaurantParser.toRestaurant(new JsonObject()); 
```

project/backend/src/test/java/com/google/sps/JsonToRestaurantParserTest.java
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  @Test(expected = JSONException.class)
+  public void missingMandatoryField() throws JSONException {
+    JSONObject body = new JSONObject()
+      .put("place_id", PLACE_ID)
+      .put("name", RESTAURANT_NAME)
+      .put("vicinity", VICINITY)
+      .put("types", EMPTY_TYPES);
+    Restaurant actual = JsonToRestaurantParser.toRestaurant(body);
```
surbhimadan1995, line 99: nit: don't need to have actual declared as a local var -- just leaving it at JsonToRestaurantParser.toRestaurant(body);

project/backend/src/test/java/com/google/sps/JsonToRestaurantParserTest.java
```diff
@@ -0,0 +1,100 @@
...
+    assertThat(actual).isEqualTo(expected);
+  }
+
+  @Test(expected = JSONException.class)
+  public void noFields() throws JSONException {
+    Restaurant actual = JsonToRestaurantParser.toRestaurant(new JSONObject());
+  }
+
+  @Test(expected = JSONException.class)
+  public void missingMandatoryField() throws JSONException {
```
miahsanchez, line 92: which fields are mandatory?

kpancha: The ones that are in the noOptionalFields test

miahsanchez: oh so here the geometry is excluded right?

project/backend/src/test/java/com/google/sps/RestaurantScorerTest.java
```diff
@@ -0,0 +1,62 @@
...
+      PLACE_ID, RESTAURANT_NAME, VICINITY, COORDS, RATING, NUM_RATINGS, PRICE_LEVEL, ALL_TYPES);
+
+  @Test
+  public void noOptionalPrefFieldsPresent_useRatingScore() throws JSONException {
+    JSONObject preferences = new JSONObject()
+      .put("currLocation", new JSONObject().put("lat", COORDS.get("lat")).put("lng", COORDS.get("lng")));
+    double actual = RestaurantScorer.score(RESTAURANT_ALL_FIELDS, preferences, STATS);
+    int numInitialRatings = (int) Math.round(STATS.getMean() / 5);
+    double weightedRating = (numInitialRatings * MID_RATING + NUM_RATINGS * RATING) / (NUM_RATINGS + numInitialRatings);
+    double expected = (weightedRating - MID_RATING) / (MAX_RATING - MID_RATING);
```
kpancha, line 44: Maybe this should be stored in a static final var RATING_SCORE so that we don't have to do the same calculation in every test

project/backend/src/test/java/com/google/sps/RestaurantScorerTest.java
```diff
@@ -0,0 +1,62 @@
...
+    assertThat(actual).isEqualTo(expected);
+  }
+
+  @Test
+  public void allPrefFieldsPresentAndMatching() throws JSONException {
+    JSONObject preferences = new JSONObject()
+      .put("currLocation", new JSONObject().put("lat", COORDS.get("lat")).put("lng", COORDS.get("lng")))
+      .put("priceLevel", new JSONObject().put("pref", PRICE_LEVEL).put("weight", 2))
+      .put("diningExp", new JSONObject().put("pref", "meal_takeaway").put("weight", 3))
+      .put("radius", new JSONObject().put("pref", 10000).put("weight", 4));
```
kpancha, line 54: Thinking of storing the weights and radius in static final vars like WEIGHT_2, WEIGHT_3, etc. for readability?

project/backend/src/test/java/com/google/sps/RestaurantScorerTest.java
```diff
@@ -0,0 +1,62 @@
...
+    double actual = RestaurantScorer.score(RESTAURANT_ALL_FIELDS, preferences, STATS);
+    int numInitialRatings = (int) Math.round(STATS.getMean() / 5);
+    double weightedRating = (numInitialRatings * MID_RATING + NUM_RATINGS * RATING) / (NUM_RATINGS + numInitialRatings);
+    double ratingScore = (weightedRating - MID_RATING) / (MAX_RATING - MID_RATING);
+    double expected = (ratingScore + 9) / 10;
```
kpancha, line 59: That way, if we have the weights already stored we can sum them up in this line instead of hard coding the values

project/backend/src/test/java/com/google/sps/RestaurantScorerTest.java
surbhimadan1995: yep - if going to be used in multiple tests, def a static final var

project/backend/src/test/java/com/google/sps/JsonToRestaurantParserTest.java
miahsanchez: marking as resolved (discussed in pod sync): I didn't think this test was necessary since this is something that would never actually happen!

project/backend/src/test/java/com/google/sps/RestaurantScorerTest.java
```diff
@@ -0,0 +1,62 @@
...
+
+  @Test
+  public void noOptionalPrefFieldsPresent_useRatingScore() throws JSONException {
+    JSONObject preferences = new JSONObject()
+      .put("currLocation", new JSONObject().put("lat", COORDS.get("lat")).put("lng", COORDS.get("lng")));
+    double actual = RestaurantScorer.score(RESTAURANT_ALL_FIELDS, preferences, STATS);
+    int numInitialRatings = (int) Math.round(STATS.getMean() / 5);
+    double weightedRating = (numInitialRatings * MID_RATING + NUM_RATINGS * RATING) / (NUM_RATINGS + numInitialRatings);
+    double expected = (weightedRating - MID_RATING) / (MAX_RATING - MID_RATING);
+    assertThat(actual).isEqualTo(expected);
```
surbhimadan1995, line 45: also nit: I know the meeting times example used just actual and expected as names but in general, we should make these more descriptive -- actualScore, expectedScore etc 
(as in its nice to keep actual/expected to disambiguate the two, but also a suffix that describes what it is )

project/backend/src/test/java/com/google/sps/RestaurantScorerTest.java
surbhimadan1995: 👍  if used in multiple tests , otherwise, local named variables are good too if its just for this one test

project/backend/src/test/java/com/google/sps/RestaurantScorerTest.java
```diff
@@ -0,0 +1,62 @@
...
+  private static final String PLACE_ID = "A12345";
+  private static final String VICINITY = "10 Main Street";
+  private static final int MAX_RATING = 5;
+  private static final int MID_RATING = 3;
+  private static final Map<String, Double> COORDS = new HashMap<String, Double>() {{
+    put("lat", 10.0);
+    put("lng", 20.0);
+  }};
+  private static final DescriptiveStatistics STATS = new DescriptiveStatistics(new double[] {10, 30, 50, 60, 90, 100});
+  private static final Restaurant RESTAURANT_ALL_FIELDS = Restaurant.create(
```
surbhimadan1995, line 34: if the arguments going into .create here won't be used individually in tests, then fine to inline them with /* paramTags */ as needed

project/backend/src/test/java/com/google/sps/JsonToRestaurantParserTest.java
```diff
@@ -0,0 +1,100 @@
...
+    Restaurant expected = Restaurant.create(
+      PLACE_ID, RESTAURANT_NAME, VICINITY, COORDS, /* avgRating= */ MISSING_FIELD, 
+      /* numRatings= */ MISSING_FIELD, /* priceNum= */ MISSING_FIELD, EMPTY_TYPES);
+    Restaurant actual = JsonToRestaurantParser.toRestaurant(body);
+    assertThat(actual).isEqualTo(expected);
+  }
+
+  @Test(expected = JSONException.class)
+  public void noFields() throws JSONException {
+    Restaurant actual = JsonToRestaurantParser.toRestaurant(new JSONObject());
```
surbhimadan1995, line 88: nit: skip saving the variable as its not used

project/backend/src/test/java/com/google/sps/RestaurantScorerTest.java
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  @Test
+  public void allPrefFieldsPresentAndMatching() throws JSONException {
+    JSONObject preferences = new JSONObject()
+      .put("currLocation", new JSONObject().put("lat", COORDS.get("lat")).put("lng", COORDS.get("lng")))
+      .put("priceLevel", new JSONObject().put("pref", PRICE_LEVEL).put("weight", WEIGHT_2))
+      .put("diningExp", new JSONObject().put("pref", "meal_takeaway").put("weight", WEIGHT_3))
+      .put("radius", new JSONObject().put("pref", 10).put("weight", WEIGHT_4));
+    double actualScore = RestaurantScorer.score(RESTAURANT_ALL_FIELDS, preferences, STATS);
```
surbhimadan1995, line 61: nit: add blank line between set up and testing (so right before actualScore) in all tests -- just like you've done in JsonToRestaurantParserTest

project/backend/src/test/java/com/google/sps/RestaurantScorerTest.java
```diff
@@ -0,0 +1,133 @@
...
+      prefs = new JSONObject()
+        .put("currLocation", new JSONObject().put("lat", COORDS.get("lat")).put("lng", COORDS.get("lng")))
+        .put("priceLevel", new JSONObject().put("pref", PRICE_LEVEL).put("weight", WEIGHT_2))
+        .put("diningExp", new JSONObject().put("pref", "meal_takeaway").put("weight", WEIGHT_3))
+        .put("radius", new JSONObject().put("pref", 10).put("weight", WEIGHT_4));
+    } catch (JSONException e) {
+      throw new ExceptionInInitializerError(e);
+    }
+    PREFS_ALL_FIELDS = prefs;
+  }
```
jesseelizabeth, line 59: I think you could just write this as:
`private static final JSONObject PREFS_ALL_FIELDS = new JSONObject().put(.....);`

jesseelizabeth: Synced offline about this - may try adding a before block for this

project/backend/src/test/java/com/google/sps/RestaurantScorerTest.java
```diff
@@ -43,19 +44,15 @@
...
-    PREFS_ALL_FIELDS = prefs;
+  private static final JSONObject PREFS_ALL_FIELDS = new JSONObject();
+
+  @Before
+  public void setPrefs() throws JSONException {
+    PREFS_ALL_FIELDS
+      .put("currLocation", new JSONObject().put("lat", COORDS.get("lat")).put("lng", COORDS.get("lng")))
+      .put("priceLevel", new JSONObject().put("pref", PRICE_LEVEL).put("weight", WEIGHT_2))
+      .put("diningExp", new JSONObject().put("pref", "meal_takeaway").put("weight", WEIGHT_3))
+      .put("radius", new JSONObject().put("pref", 10).put("weight", WEIGHT_4));
```
jesseelizabeth, line 55: Looks great!

PR #56, opened 2020-07-25
Style PreferencesForm component

alexcostaluiz:
Styles the `PreferenceForm` component and changes it from a class component to a functional one.

image


project/frontend/src/pages/HomePage.js
```diff
@@ -93,12 +93,9 @@ function HomePage(props) {
...
       top='64px'
-      bottom='64px'>
-      <PreferenceForm
-        history={props.history}
-        currLocation={currLocation}
-        locationName={locationName}
-      />
+      bottom='64px'
+      style={{ paddingBottom: '0px' }}>
+      <PreferenceForm currLocation={currLocation} locationName={locationName} />
```
jesseelizabeth, line 98: Is history not needed anymore?

alexcostaluiz: It is still needed, but I decided to use the `useHistory` hook instead of passing history down as a prop. Since the component is now a functional one and can use hooks.

PR #57, opened 2020-07-27
Style location finder

alexcostaluiz:
Style the `LocationFinder` component and add a logo to the `HomePage` and `Header` components.

styled_location_finder


project/frontend/src/components/LocationFinder.js
```diff
@@ -1,5 +1,8 @@
+import './LocationFinder.css';
+
 import React from 'react';
-import MyLocation from '@material-ui/icons/MyLocation';
+import Loading from './Loading.js';
+import MyLocation from '../assets/gps.svg';
```
kpancha, line 5: Does this mean that you just saved the MyLocation icon locally in our project?

alexcostaluiz: Yeah, I don't remember if its the same icon, but I wanted more control over the styling and stuff.

project/frontend/src/pages/HomePage.js
```diff
@@ -20,20 +20,29 @@ function HomePage(props) {
     setLocationName(locationData.locationName);
   };
 
+  const buttonStyle = {
+    opacity: locationName ? 1 : 0,
+    transform: locationName ? 'translateY(0px)' : 'translateY(-24px)',
+  };
+
```
surbhimadan1995, line 27: optional: if more stuff gets added to this, would it be cleaner to have 
```
const buttonStyle = locationName ? { opacity: 1, transform: 0px} : { opacity: 0, transform: -24px} 
```
(if thats a syntax thats valid)

alexcostaluiz: Tbh I don't anticipate any more styles being added here since I already have a class on these buttons that sets their constant styles. I'll swap it out for your recommendation though because I do think the one ternary is cleaner than two. Thanks!

PR #59, opened 2020-07-28
Implement swipe match page

alexcostaluiz:
Ready for review.

- Implements the `SwipeMatchPage` component
- Adds a `FirestoreContext` which provides the db instance to any component
- Adds a `SwipeMatchService` class responsible for making raw commands to the database
- Refactors `RestaurantCardDeck` so a listener for swipe events may be provided as a prop

Screenshot 2020-08-05 at 3 23 57 PM


project/backend/src/main/java/com/google/sps/servlets/GenerateSwipeMatchGroupIdServlet.java
```diff
@@ -0,0 +1,89 @@
...
+  public void init() {
+    try {
+      firestoreOptions = FirestoreOptions.getDefaultInstance().toBuilder()
+        .setProjectId(PROJECT_ID)
+        .setCredentials(GoogleCredentials.getApplicationDefault())
+        .build();
+      
+      db = firestoreOptions.getService();
+    } catch (IOException e) {
+      // Not sure what to do if initialization fails...
```
surbhimadan1995, line 44: hmmm, this would mean we can't use the db for handling requests, so we could make it Optional<Firestore> (https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html ) db and if we hit an IOException, initialize it to Optional.empty -- then on doGet -- we can send an internal server error in the response if the db doesn't exist

project/backend/src/main/java/com/google/sps/servlets/GenerateSwipeMatchGroupIdServlet.java
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  private boolean isGroupIdUnique(String groupId) {
+    try {
+      return !db.collection(COLLECTION_NAME).document(groupId).get().get().exists();
+    } catch (InterruptedException e) {
+      return false;
+    } catch (ExecutionException e) {
+      return false;
+    }
```
surbhimadan1995, line 87: nit: we can merge these catch blocks into one as they're handled in the same way 
catch (InterruptedException | ExecutionException e)

project/frontend/src/components/SwipeMatchConsole.js
```diff
@@ -0,0 +1,83 @@
...
+
+  return (
+    <div className='swipe-match-console'>
+      <h4>Swipe Match</h4>
+      <p>
+        Welcome to Swipe Match. Below you will find a group ID to share with
+        others so they may join the session.
+        {isCreator
+          ? ' As the session leader, you may moderate the connected users and are responsible for starting the session once restaurants have been retrieved. You may start the session at any time, and users may still join after you have begun the session.'
+          : ' Please wait for the session leader to begin the session.'}
```
surbhimadan1995, line 22: nit: break into reasonable length lines (maybe roughly as long as line 18)

project/backend/src/main/java/com/google/sps/servlets/GenerateSwipeMatchGroupIdServlet.java
```diff
@@ -0,0 +1,89 @@
+package com.google.sps.servlets;
```
surbhimadan1995, line 1: Since we added junit tests for servlets, we could add some for this too?

project/frontend/src/pages/SwipeMatchPage.js
```diff
@@ -0,0 +1,185 @@
...
+          let user = authContext.currentUser.get;
+
+          if (user && user.isSignedIn()) {
+            let creatorCurrentSwipeMatchSession = null;
+            try {
+              creatorCurrentSwipeMatchSession = await SwipeMatchService.fetchCreatorCurrentSwipeMatchSession(
+                firestore,
+                user
+              );
+            } catch (e) {}
```
surbhimadan1995, line 83: if its intentional to do nothing on catching the error here, then lets leave a comment with // do nothing, otherwise a TODO on how we'd handle it

project/frontend/src/pages/SwipeMatchPage.js
```diff
@@ -0,0 +1,185 @@
...
+        }
+      };
+    },
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    isCreator ? [authContext] : []
+  );
+
+  useEffect(() => {
+    if (location.lat && location.lng) {
+      console.log('ACTUALLY UPDATING RESTAURANTS');
```
surbhimadan1995, line 132: seems unnecessary?

project/frontend/src/pages/SwipeMatchPage.js
alexcostaluiz: Ended up refactoring this whole section out, so resolved!

PR #60, opened 2020-07-29
Profile Page React

miahsanchez:
front end for the profile page!

This is what the profile page currently looks like there are some styling things I want to do / add but this is what I'm thinking for all parts of the profile page that the user will interact with:
Screen Shot 2020-07-31 at 10 17 40 AM


project/frontend/src/components/ProfileForm.js
```diff
@@ -0,0 +1,178 @@
...
+import Place from '../assets/place.svg';
+import Cuisine from '../assets/cuisine.svg';
+import Distance from '../assets/distance.svg';
+import Experience from '../assets/food_service.svg';
+import Price from '../assets/dollar.svg';
+
+/**
+ * A form used to gather the restaurant preferences of a user for the 'Find My Match'
+ * service.
+ *
```
surbhimadan1995, line 19: er... a lot of this code is the same as the preference form component: https://github.com/googleinterns/step127-2020/blob/master/project/frontend/src/components/PreferenceForm.js ? is that intentional or can it be shared using props ?

miahsanchez: Yup right now it's copied but I don't plan on keeping it like this!! copied from my comment upon opening the PR: "opening this draft because i want opinions about how i should make components from the preference form page so that it can be used on the profile page." this morning Kira, Alex, and I were discussing how to split up the components in the preference form page so that they can be used here.

I wanted to show them what code would be necessary to include in the profile page to help brainstorm how we could create components that would work for both the preference form and the profile page.

surbhimadan1995: ah ok, sounds good! sorry for hopping on preemptively :D

project/frontend/src/components/ProfileForm.js
```diff
@@ -0,0 +1,69 @@
...
+  const [cuisine, setCuisine] = useState([]);
+
+  return (
+    <div className='profile-form-container'>
+      <form className='profile-form'>
+        <h4 className='profile-header'>Your profile.</h4>
+        <div className='profile-form-row'>
+          <img src={Place} alt='' />
+          <label>Your Location</label>
+          {/* TODO: change so that if a location isn't inputted then we are going to need to have a fill in optiob. */}
```
kpancha, line 33: We could just add a LocationFinder here with sendData={handleLocationData} where handleLocationData sets the state or does whatever we want do do with the location.

project/frontend/src/components/ProfileForm.js
```diff
@@ -0,0 +1,69 @@
...
+          <p>Your location</p>
+        </div>
+        <div className='profile-form-row'>
+          <img src={Cuisine} alt='' />
+          <label htmlFor='cuisine'>Preferred Cuisines</label>
```
kpancha, line 38: We use the htmlFor to associate a label with an html element. So whatever this is labeling needs to have `id='cuisine'` so that you know what it is to be associated with and same with all the ones below (distance, price, experience). Like in this case, I think we'd want to make 'cuisine' the id for the CuisineAutocomplete component. The responses here explain this nicely: https://stackoverflow.com/questions/698001/whats-the-for-for-in-a-label-tag

project/frontend/src/components/ProfileForm.js
```diff
@@ -0,0 +1,69 @@
...
+          <CuisineAutocomplete
+            cuisineOptions={cuisineOptions}
+            setCuisine={setCuisine}
+          />
+        </div>
+        <div className='profile-form-row'>
+          <img src={Distance} alt='' />
+          <label htmlFor='distance'>Ideal Distance</label>
```
kpancha, line 46: Going off what I commented above, we could maybe be a little more specific with the id as well? So this could be `htmlFor='distance-button-group'` and then below the ButtonGroup component could have `id='distance-button-group'`. Just a thought

project/frontend/src/components/ProfileForm.js
```diff
@@ -0,0 +1,69 @@
...
+function ProfileForm(props) {
+  const cuisineOptions = ['Italian', 'Mexican', 'Indian'];
+  const [cuisine, setCuisine] = useState([]);
+
+  return (
+    <div className='profile-form-container'>
+      <form className='profile-form'>
+        <h4 className='profile-header'>Your profile.</h4>
+        <div className='profile-form-row'>
+          <img src={Place} alt='' />
```
kpancha, line 31: alt here can probably be something simple like 'place-icon'. same goes for the other icons below

project/frontend/src/components/ProfileForm.js
```diff
@@ -0,0 +1,69 @@
...
+        </div>
+        <div className='profile-form-row'>
+          <img src={Price} alt='' />
+          <label htmlFor='price'>Price Level</label>
+          <ButtonGroup labelList={['$', '$$', '$$$', '$$$$']} />
+        </div>
+        <div className='profile-form-row'>
+          <img src={Experience} alt='' />
+          <label htmlFor='experience'>Dining Experience</label>
+          <ButtonGroup labelList={['Takeout', 'Delivery', 'Dine-In']} />
```
kpancha, line 59: Not something to change now, but I think this could easily be turned into a map to map each label to its corresponding value, and then we just handle a map instead of a list in ButtonGroup.  If we get that far next week😅

project/frontend/src/components/ProfileForm.js
miahsanchez: kk. i wasnt sure bc there weren't any alt values in the preference form

project/frontend/src/components/ProfileForm.js
miahsanchez: gotcha! should i also change this in the preference form then? I was just following that structure

kpancha: We can open a new PR to make those changes on PreferenceForm (adding alt text, fixing labels) :)

kpancha: Also regarding labels, I think we talked about this way earlier but instead of using for you can also wrap the thing that is being labeled in the label tag if you want.. I'm pretty sure you had used it that way for something earlier, so that would work as well since it still binds the label to the element (under "Tips and Notes") https://www.w3schools.com/tags/tag_label.asp

project/frontend/src/components/ProfileForm.js
kpancha: yep we should add those!

jesseelizabeth:
Is this PR going to be closed and replaced with PR #68 ?

miahsanchez:
> Is this PR going to be closed and replaced with PR #68 ?

I was planning on closing this one yes! I just wanted to take a look at these comments and update the other PR based on them before closing this one!

PR #61, opened 2020-07-29
Fix Map Styling

alexcostaluiz:
I've just copied the changes we made to the `infoWindow` components and the map styling before our demo to this branch so we can merge these changes into master.

Also renamed ResultsMap.js -> MapContainer.js.


project/frontend/src/components/MapContainer.js
```diff
@@ -65,30 +67,30 @@ function MapContainer(props) {
...
     };
 
     return (
-      <div style={infoWindowStyle}>
-        <div style={{ fontSize: 16 }}>
-          <span style={{ float: 'right' }}>{percentMatch}</span>
-          {restaurantName}
-        </div>
+      <div className='restaurant-header' style={infoWindowStyle}>
+        <h5 className={'restaurant-name'}>{restaurantName}</h5>
```
miahsanchez, line 83: is there a reason this is in brackets?

alexcostaluiz: Oh no there is not, i'll fix real quick

project/frontend/src/components/MapContainer.js
```diff
@@ -65,30 +67,30 @@ function MapContainer(props) {
...
+        <h5 className='restaurant-score'>{percentMatch}</h5>
```
miahsanchez, line 84: is it a convention to add classNames to everything? I'm just wondering if we need them if we don't have a css file for this

alexcostaluiz: The class names do actually reference a class I have defined in `RestaurantCard.css` so that they match the header in the card.

alexcostaluiz: Since this map is always being used with the cards nearby I figured it was safe to assume that `RestaurantCard.css` would be loaded.

miahsanchez: so you don't have to import `RestaurantCard.css` into `MapContainer.js` in order for the styling to work?

alexcostaluiz: No not explicitly. When you import a stylesheet react just appends the entire sheet's content inside the <head> element of the html page, so when it was imported in RestaurantCard.js it got added to <head> and then when this map renders it finds that css class in the <head> element.

PR #63, opened 2020-07-30
Minor algorithm tweaks

kpancha:
- Adds tooltip to sliders (see picture below)
- limits number of cuisines that user can choose to 10
Screenshot 2020-07-31 at 11 54 22 AM


project/frontend/src/components/PreferenceForm.js
```diff
@@ -105,7 +111,7 @@ function PreferenceForm(props) {
 
   const getSlider = (value, setValue, disabled) => {
     return (
-      <div className='preference-form-slider-container'>
+      <div className='preference-form-slider-container' style={{ paddingLeft: 15}}>
```
alexcostaluiz, line 114: Could we move the `paddingLeft` style here to the `preference-form-slider-container` class?

project/frontend/src/components/CuisineAutocomplete.js
```diff
@@ -19,7 +19,9 @@ import TextField from '@material-ui/core/TextField';
  */
 function inputIsNewAndValid(input, options) {
   const trimmedInput = input.trim();
+  const maxNumCuisines = 5;
```
alexcostaluiz, line 22: I think we should do 10 instead of 5

project/frontend/src/components/PreferenceForm.js
```diff
@@ -71,6 +71,12 @@ function PreferenceForm(props) {
     })();
   }, [currLocation]);
 
+  useLayoutEffect(() => {
+    setTooltipVal(radiusWeight, 0);
+    setTooltipVal(diningExpWeight, 1);
+    setTooltipVal(priceLevelWeight, 2);
+  }, [radiusWeight, diningExpWeight, priceLevelWeight]);
+
```
alexcostaluiz, line 79: I feel like doing something with a loop would be better for future-proofing this code. Idk if we'll actually get around to it, but if we were to add more preferences, this method of doing it could get really verbose.

Event simply passing an array to `setTooltipVal` I think would be a lot better. For example:
```
setTooltipVal([radiusWeight, diningExpWeight, priceLevelWeight]);
```
and then inside `setTooltipVal`:
```
for (const tooltip of tooltips) {
  tooltips[tooltipInd].innerHTML = levels[val - 1];
}
```

project/frontend/src/components/PreferenceForm.js
```diff
@@ -120,6 +126,21 @@ function PreferenceForm(props) {
     );
   };
 
+  const setTooltipVal = (val, tooltipInd) => {
```
alexcostaluiz, line 129: If you go with the array param route I talked about above, this is not applicable, but i feel like `tooltipInd` should be either `index`, `tooltipIdx`, or `tooltipIndex`. Idk if this is just me but `Ind` feels like a weird abbreviation of index.

project/frontend/src/components/PreferenceForm.js
kpancha: This is a great idea! Done in latest commit (a90ad05)

project/frontend/src/components/CuisineAutocomplete.js
```diff
@@ -19,7 +19,9 @@ import TextField from '@material-ui/core/TextField';
  */
 function inputIsNewAndValid(input, options) {
   const trimmedInput = input.trim();
+  const maxNumCuisines = 10;
   return (
+    options.length < maxNumCuisines &&
```
surbhimadan1995, line 24: does this mean if we get fewer than 10 cuisines that are available that we can list, this would be invalid input?

kpancha: oh oops yes you're right I need to fix this to use the selected options

surbhimadan1995: 👍 yeah it should be display max(numCuisines, 10) options

project/frontend/src/components/PreferenceForm.js
```diff
@@ -168,7 +190,8 @@ function PreferenceForm(props) {
           </div>
         )}
       <form className='preference-form' onSubmit={handleSubmit}>
-        <h4>Your preferences.</h4>
+        <h4>Your preferences. <Tooltip title='hello world'><Info /></Tooltip></h4>
+        <div className='tooltip'>hello world</div>
```
miahsanchez, line 194: this is filler text right?

kpancha: Yep I am going to make a real tooltip and insert it here :)

project/frontend/src/components/CuisineAutocomplete.js
```diff
@@ -17,7 +17,7 @@ import TextField from '@material-ui/core/TextField';
  * @param {string} input The input string to be considered.
  * @param {!Array<string>} options The complete list of cuisine options.
  */
-function inputIsNewAndValid(input, options) {
+function inputIsNewAndValid(input, options, numSelected) {
```
miahsanchez, line 20: Why do you pass in numSelected if you don't use it in this function?

project/frontend/src/components/CuisineAutocomplete.js
```diff
@@ -39,32 +39,40 @@ function inputIsNewAndValid(input, options) {
...
   const filter = createFilterOptions();
+  const maxNumCuisines = 10;
 
   return (
     <Autocomplete
       name='cuisine'
       id='cuisine'
       options={cuisineOptions}
       multiple
+      limitTags={10}
```
miahsanchez, line 53: could this be maxNumCuisines instead of 10 so this isn't a magic number?

project/frontend/src/components/PreferenceForm.js
```diff
@@ -128,18 +145,40 @@ function PreferenceForm(props) {
...
+    'Low': 1,
+    'Medium': 2,
+    'High': 3,
     'Very High': 4,
   };
 
+  const tooltipInfo = 
+  (<React.Fragment>
+    <p>We will use your preferences entered here to find a restaurant we think you'd like.</p>
+    <p>You may leave any field blank if you have no preference for that field. If you are looking for a
```
miahsanchez, line 163: i think you could do a line break `<br>` instead of having a few different <p> elements unless there's a reason these need to be separated.

kpancha: Although it would visually look the same, I used different paragraphs because they're like separate thoughts, whereas a line break is to manually break up a line within a paragraph... see https://stackoverflow.com/questions/13688158/when-to-use-p-vs-br#:~:text=You%20want%20to%20use%20the,of%20information%20into%20separate%20thoughts.&text=The%20tag%20is,line%2C%20such%20as%20with%20poetry.&text=Hello%20World!

PR #66, opened 2020-07-31
Button on click functionality

miahsanchez:
draft pr for the onClick functionality of buttons on the profile page. All that's been implemented so far is the button background color changes when any button is clicked on.


project/frontend/src/components/ButtonGroup.js
```diff
@@ -9,11 +9,38 @@ import React from 'react';
  */
 function ButtonGroup(props) {
   const { labelList } = props;
-  const buttonList = [];
-  for (let delta = 0; delta < labelList.length; delta++) {
-    buttonList.push(<button>{labelList[delta]}</button>);
-  }
-  return <div className='btn-group'>{buttonList}</div>;
+  const [backgroundIsGray, setBackgroundIsGray] = useState({
```
kpancha, line 12: Maybe we could rename this to `isSelected` or something (it looks to me like that is what backgroundIsGray indicates but correct me if I'm wrong!) just so it is clearer what is happening

project/frontend/src/components/ButtonGroup.js
```diff
@@ -9,11 +9,38 @@ import React from 'react';
...
+
+  const onClickButton = (props) => {
+    let tempBackgroundIsGray = Object.assign({}, backgroundIsGray);
+    tempBackgroundIsGray[props] = !tempBackgroundIsGray[props];
+    setBackgroundIsGray(tempBackgroundIsGray);
+  };
+
+  const createButtons = () => {
+    const buttonList = [];
+    for (let delta = 0; delta < labelList.length; delta++) {
```
kpancha, line 27: Sorry to be so nit about all the naming of stuff lol, but I'm not sure if delta is the best name here? I'm guessing this is based on a different PR where we used delta to describe the offset from 0, but I think here we could just use `i` or `buttonIndex` or something along those lines?

project/frontend/src/components/ButtonGroup.js
```diff
@@ -9,11 +9,38 @@ import React from 'react';
...
+      let className = backgroundIsGray[delta]
```
kpancha, line 28: Probably should check for index out of bounds in case labelList is longer that backgroundIsGray. And/or the initialization of backgroundIsGray could be based on however long labelList is?

miahsanchez: yup I was going to do this later today! I need to change the way that I initialize backGroundIsGray because it isn't dynamic

kpancha:
Looks great and nice specific variable names :)

PR #68, opened 2020-08-03
New Preference form component

miahsanchez:
I made a new preference form component `PreferenceForm.js` to try and minimize repeated code between `UserPreferenceForm.js` and `ProfileForm.js`

Changes:

**PreferenceForm.js** A new form component that will handle user input for cuisines, price, distance, and dining experience.

**FormItem.js** this is for a single input item in a form. It creates the label and image for an input item and also renders its children.

I have updated UserPreferenceForm.js to use the `PreferenceForm` component and the functionality of this form is still the same!


project/frontend/src/components/PreferenceForm.js
```diff
@@ -1,251 +1,18 @@
...
-          <input
-            name='open'
-            type='checkbox'
-            checked={open}
-            onChange={(event) => setOpen(event.target.checked)}
-          />
-        </div>
-        <div className='preference-form-submit-container'>
-          <button type='submit'>Find my match</button>
+    <div className='form-container'>
```
miahsanchez, line 7: add the cuisine, distance, etc here

project/frontend/src/components/PreferenceForm.js
```diff
@@ -1,251 +1,18 @@
...
+      <form className='form'>
```
miahsanchez, line 8: and add a prop for the labeling

project/frontend/src/components/UserPreferenceForm.js
```diff
@@ -214,86 +226,41 @@ function PreferenceForm(props) {
...
+        blank if you have no preference. Specify an importance to indicate your
+        priority for different fields.
+      </p>
+      <PreferenceForm
+        rowItemLabels={{
+          cuisine: 'Cuisines',
+          location: 'Location',
+          experience: 'Experience',
+          price: 'Price Level',
+          distance: 'Distance',
```
jesseelizabeth, line 246: nit: put these in a variable with the other vars (distancesInMiles, prices etc) above

project/frontend/src/components/PreferenceForm.js
```diff
@@ -0,0 +1,87 @@
...
+                <label>Importance</label>
+                {props.children[5]}
+              </div>
+            </div>
+          )}
+        </div>
+        {isUserPreference && props.children[6]}
+        <div className='form-submit-container'>
+          <button type='submit'>{buttonLabel}</button>
+        </div>
```
jesseelizabeth, line 82: Since these are only shown in the UserPreferenceForm maybe we should just leave them there. Is there a reason we need them in here?

miahsanchez: If I dont include them inside the `<PreferenceForm>` tags they show up under the submit button so the form looks wonky

jesseelizabeth: Ah I see. Sg to leave here then!

project/frontend/src/components/FormItem.css
```diff
@@ -0,0 +1,30 @@
+.row {
+  display: flex;
+  align-items: center;
+  margin: 16px 0px;
+}
```
alexcostaluiz, line 5: We should make these class names more specific. Unfortunately importing styles in React isn't as modular as it would seem. Every class we make just gets thrown into the `<head>` section of the html page. So any classnames that are the same will overwrite each other depending on which gets imported last.

What I've been doing is basing class names on the component name, so like `form-item-row` would be better for this.

project/frontend/src/components/PreferenceForm.css
```diff
@@ -0,0 +1,77 @@
...
+}
+
+.form-container button {
+  display: block;
+  margin: 16px auto;
+}
+
+.form {
+  margin-bottom: 0px;
+}
```
alexcostaluiz, line 14: form, form-container, and header are also a bit generic since we have a lot of forms in this app. I think `preference-form-...` would be better.

project/frontend/src/components/ProfileForm.css
```diff
@@ -0,0 +1,16 @@
...
+.profile-form-container button {
+  display: block;
+  margin: 16px auto;
+}
+
+.header {
+  text-align: center;
+  margin-bottom: 20px;
+  margin-top: 15px;
+}
```
alexcostaluiz, line 16: same for this one

project/frontend/src/components/PreferenceForm.js
```diff
@@ -0,0 +1,87 @@
...
+        <FormItem imageName={Cuisine} label={rowItemLabels.cuisine}>
+          <CuisineAutocomplete
+            cuisineOptions={cuisineOptions}
+            setCuisine={setCuisine}
+          />
+        </FormItem>
+        <div className='form-row'>
+          <div className='form-column'>
+            <FormItem imageName={Distance} label={rowItemLabels.distance}>
+              {props.children[0]}
```
kpancha, line 53: What are these specific indeces from props.children/is there a way to make it less hard codey to prevent index out of bounds?

miahsanchez: Unfortunately no (for now). I've discussed alternatives with Jesse but this seems to be the best way for now. Luckily an out of bounds error wont happen when the user side just on the developer side if someone were to change props.children. Ive added documentation accounting for this since it isn't ideal:
https://github.com/googleinterns/step127-2020/blob/cd2125a364c3b020d0609df7e2127dfb0a637e19/project/frontend/src/components/ProfileForm.js#L36

PR #69, opened 2020-08-04
Check open by default if time is between 10am and 10pm

kpancha:


project/frontend/src/components/PreferenceForm.js
```diff
@@ -52,7 +52,10 @@ async function getLocalCuisines(currLocation) {
  */
 function PreferenceForm(props) {
   const { currLocation, locationName } = props;
-
+  const currHour = parseInt(
+    new Date(Date.now()).toTimeString().substring(0, 2)
```
miahsanchez, line 56: if it's 22:30 would currHour be 22 or 21? just asking because we wouldn't want openNow to be checked at 22:30

kpancha: Oh good point-- it should be [10,22) not [10,22]

project/frontend/src/components/PreferenceForm.js
```diff
@@ -52,7 +52,10 @@ async function getLocalCuisines(currLocation) {
...
+  );
+  const isOpenCheckedDefault = currHour >= 10 && currHour <= 22;
```
surbhimadan1995, line 58: probably worth leaving s comment here with just "open now is checked by default if the current time is between 10am and 10pm" -- the >= 10 && <= 22 check might not give that info at just first glance

PR #71, opened 2020-08-05
Fix Undefined Property Error in Google Map Component

alexcostaluiz:
Small fix for the Google map component.

### DIFF
--- a/project/frontend/src/components/MapContainer.js
+++ b/project/frontend/src/components/MapContainer.js
@@ -109,6 +109,7 @@ function MapContainer(props) {
         const coords = restaurants[addMarkerIndex].key.latLngCoords;
         markers.push(
           <MarkerIcon
+            key={restaurants[addMarkerIndex].key.id}
             lat={coords.lat}
             lng={coords.lng}
             id={addMarkerIndex}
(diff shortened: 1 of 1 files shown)


miahsanchez:
since you added keys to all the children could you add one to the info window?

alexcostaluiz:
> since you added keys to all the children could you add one to the info window?
.
InfoWindow has no need for a key. They're only relevant for components in an array. If an array of components every changes (items removed, added, or index changes), keys let react know how to preserve state for the right components. Since InfoWindow is not part of an array of siblings it doesn't need a key.

miahsanchez:
> > since you added keys to all the children could you add one to the info window?
> 
> InfoWindow has no need for a key. They're only relevant for components in an array. If an array of components every changes (items removed, added, or index changes), keys let react know how to preserve state for the right components. Since InfoWindow is not part of an array of siblings it doesn't need a key.

so then why does the you are here marker need one?

alexcostaluiz:
> > > since you added keys to all the children could you add one to the info window?
> > 
> > 
> > InfoWindow has no need for a key. They're only relevant for components in an array. If an array of components every changes (items removed, added, or index changes), keys let react know how to preserve state for the right components. Since InfoWindow is not part of an array of siblings it doesn't need a key.
> 
> so then why does the you are here marker need one?

Ah yea I meant to mention, the Google Map uses keys of any marks during the whole distanceFromMouse thing so it seemed important that all markers have a key even though YouAreHere is not in an array.

PR #76, opened 2020-08-06
Fix Restaurant Card Component

alexcostaluiz:
Some minor fixes to the `RestaurantCard` component.

- Do not display percent match if no percent match exists (during swipe match sessions)
- Avoid undefined pointer error if no images are returned from the Google Places API

### DIFF
--- a/project/frontend/src/components/RestaurantCard.js
+++ b/project/frontend/src/components/RestaurantCard.js
@@ -163,7 +163,9 @@ function RestaurantCard(props) {
   let photoUrls;
   if (details) {
     if (details.result.photos) {
-      photoUrls = details.result.photos.map((photo) => photo.getUrl({ maxWidth: 344 }));
+      photoUrls = details.result.photos.map((photo) =>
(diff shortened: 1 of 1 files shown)


project/frontend/src/components/RestaurantCard.js
```diff
@@ -159,9 +159,20 @@ function RestaurantCard(props) {
...
+    if (details.result.photos) {
+      photoUrls = details.result.photos.map((photo) =>
+        photo.getUrl({ maxWidth: 344 })
+      );
+    } else {
+      // TODO: Display a better "no images found" type image
+      photoUrls = ['https://pixelpapa.com/wp-content/uploads/2018/11/26.gif'];
+    }
+  } else {
+    photoUrls = ['https://pixelpapa.com/wp-content/uploads/2018/11/26.gif'];
```
surbhimadan1995, line 174: lets make the 'https://pixel...' part a const named string above maybe like placeholderImageUrl = 'http://pixel..' 
and use it here

PR #77, opened 2020-08-06
Add User Creation to the Authentication Component

alexcostaluiz:
Creates user profiles in our Firestore backend on first-time user authentications.

### DIFF
--- a/project/frontend/src/components/Authentication.js
+++ b/project/frontend/src/components/Authentication.js
@@ -63,6 +63,11 @@ function Authentication(props) {
           email: profile.getEmail(),
           imageUrl: profile.getImageUrl(),
           currentSwipeMatchSession: null,
+          location: '',
+          cuisines: [],
+          distance: [],
+          price: [],
+          experience: [],
         });
       }
     } catch (e) {
--- a/project/frontend/src/components/Authentication.js
+++ b/project/frontend/src/components/Authentication.js
@@ -47,6 +47,9 @@ function Authentication(props) {
       user = await GoogleAuth.signIn();
     } catch (e) {
       // TODO: Add component to notify user of sign in failure
+      //
+      // Ignore Google custom errors like user closing the sign-in popup
+      // without authenticating.
       return;
     }
 
@@ -72,7 +75,7 @@ function Authentication(props) {
         });
       }
     } catch (e) {
-      throw e;
+      return;
     }
   };
 
--- a/project/frontend/src/components/Authentication.js
+++ b/project/frontend/src/components/Authentication.js
@@ -20,7 +20,7 @@ const AuthContext = React.createContext({});
  */
 function Authentication(props) {
   const { firestore } = useContext(FirestoreContext);
-  
+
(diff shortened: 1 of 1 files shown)


project/frontend/src/components/Authentication.js
```diff
@@ -37,9 +41,43 @@ function Authentication(props) {
     });
   }, []);
 
+  const signInWrapper = async () => {
+    let user;
+    try {
+      user = await GoogleAuth.signIn();
+    } catch (e) {
+      return;
```
jesseelizabeth, line 49: Should we throw some sort of error here?

project/frontend/src/components/Authentication.js
```diff
@@ -37,9 +41,43 @@ function Authentication(props) {
...
+    }
+
+    const id = user.getId();
+    try {
+      const ref = firestore.collection('users').doc(id);
```
jesseelizabeth, line 54: do we have to await this as well?

alexcostaluiz: This line (the `.collection()` and `.doc()`) is completely synchronous so no need for await here. Only the `.get()`, `.update()`, `.delete()`, etc. on a document ref are asynchronous.

jesseelizabeth: Gotcha, thanks for explaining!

project/frontend/src/components/Authentication.js
alexcostaluiz: The `GoogleAuth.signIn()` function has some custom errors that it throws that aren't really indicative of a serious problem. For instance if the user closes out of the log in popup or denies access to the requested information, errors are thrown, so I figured it best to just ignore if an error is thrown.

Also throwing an error in this function would cause problems with all the unprotected `signIn()` calls we make throughout the different components. And even if we surrounded all those with try-catch, I don't think there would be anything to do really if we were to catch an error. The best we can hope for is just that the user tries to log in again.

jesseelizabeth: Do we even need a try catch then? Would be nice to have some sort of component that would indicate to the user that the signIn failed and to try again, but that can be a follow up :) Maybe for now we can just add a TODO to handle that case.

alexcostaluiz: Yeah I like the idea of a component to notify the user that it failed. I will add a TODO for now. For now the try-catch does prevent errors from being logged if the user closes the login popup without signing in.

jesseelizabeth: Sg! Let's maybe also add a comment in the catch block saying why we don't handle the error here

project/frontend/src/components/Authentication.js
```diff
@@ -37,9 +41,44 @@ function Authentication(props) {
...
+          currentSwipeMatchSession: null,
+          location: '',
+          cuisines: [],
+          distance: [],
+          price: [],
+          experience: [],
+        });
+      }
+    } catch (e) {
+      throw e;
```
kpancha: Handle this error

PR #79, opened 2020-08-07
Add Group Id Validation to SwipeMatchForm

alexcostaluiz:
- Validates group ids in the `SwipeMatchForm` component.
- Generates new group ids in the `SwipeMatchForm` component instead of the `SwipeMatchPage` component.

### DIFF
--- a/project/frontend/src/components/SwipeMatchForm.js
+++ b/project/frontend/src/components/SwipeMatchForm.js
@@ -28,6 +28,7 @@ function SwipeMatchForm(props) {
 
   const [username, setUsername] = useState('');
   const [groupId, setGroupId] = useState('');
+  const [isGroupIdValid, setIsGroupIdValid] = useState(true);
 
   const anonUsername = useRef('Anonymous ' + Animals.random());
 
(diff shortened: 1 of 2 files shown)


project/frontend/src/components/SwipeMatchForm.js
```diff
@@ -22,47 +24,114 @@ function SwipeMatchForm(props) {
...
-  const signedIn =
-    authContext.currentUser.get && authContext.currentUser.get.isSignedIn();
+
+  const user = authContext.currentUser.get;
+  const signedIn = user && user.isSignedIn();
 
   const [username, setUsername] = useState('');
   const [groupId, setGroupId] = useState('');
+  const [creatorCurrentSession, setCreatorCurrentSession] = useState(null);
+  const [isGroupIdValid, setIsGroupIdValid] = useState(true);
```
jesseelizabeth, line 36: Should we do this the other way and set this to be false initially and then set to true when it's valid?

alexcostaluiz: I can see how that seems more intuitive so maybe I should change some stuff around, but the way I'm currently using `isGroupIdValid` works better if the default is true. `isGroupIdValid` is literally only responsible for highlighting the input text box with a red border when it is flipped to true. So when a user tries to join a session, the id is validated through a backend servlet, and if its invalid, the input box gets highlighted red, then when the user focuses the input box again, it goes back to its normal border color.

If I start with false by default, the input box will be highlighted with a red border from when the popup opens to when the user first focuses it, which I think would be strange.

If `isGroupIdValid` was responsible for actually determining whether the user is able to join a session, then it should definitely be false by default, but I don't actually use it for that.

jesseelizabeth: Ah i see, wasn't sure if this was responsible for validating the group ID on the backend in any way. Let's maybe add some documentation above clarifying what this does.

project/frontend/src/components/SwipeMatchForm.js
```diff
@@ -22,47 +24,114 @@ function SwipeMatchForm(props) {
...
+          id
+        );
+      } catch (e) {}
+
+      if (id) {
+        history.push({
+          pathname: '/swipe-match',
+          state: {
+            isCreator: true,
+            username: username ? username : anonUsername.current,
```
kpancha, line 76: I believe you could just say `username || anonUsername.current` here because if username is falsy then the value will be the anon username

project/frontend/src/components/SwipeMatchForm.js
```diff
@@ -22,47 +24,114 @@ function SwipeMatchForm(props) {
...
+
+  const returnToSession = (event) => {
     event.preventDefault();
     history.push({
       pathname: '/swipe-match',
       state: {
-        currLocation,
-        action: 'create',
+        isCreator: true,
         username: username ? username : anonUsername.current,
```
kpancha, line 102: Same thing here

PR #80, opened 2020-08-07
Implement Swipe Match Header

alexcostaluiz:
Creates a `SwipeMatchHeader` component that displays the following:
- group id
- username
- connected user count
- end session / leave session button (end session button for session creators)

image

### DIFF
--- a/project/frontend/src/components/SwipeMatchHeader.js
+++ b/project/frontend/src/components/SwipeMatchHeader.js
@@ -1,0 +1,43 @@
+import './SwipeMatchHeader.css';
+
+import React from 'react';
+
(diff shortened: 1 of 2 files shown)


project/frontend/src/components/SwipeMatchHeader.js
```diff
@@ -0,0 +1,43 @@
+import './SwipeMatchHeader.css';
+
+import React from 'react';
+
+import Fingerprint from '../assets/fingerprint.svg';
+import Group from '../assets/group.svg';
+import Username from '../assets/username.svg';
+
+function SwipeMatchHeader(props) {
```
jesseelizabeth, line 9: Great use of props!

PR #81, opened 2020-08-07
Implement Swipe Match Summary

alexcostaluiz:
Creates a `SwipeMatchSummary` component which displays as a modal over the `SwipeMatchPage` when a session has ended.

image


project/frontend/src/components/SwipeMatchSummary.js
```diff
@@ -0,0 +1,80 @@
...
+                </td>
+                <td>
+                  <label>Dislikes</label>
+                </td>
+                <td>
+                  {userLikes &&
+                    users.map((user) => {
+                      if (!userLikes.includes(user)) {
+                        return <ConnectedUserChip key={user} username={user} />;
+                      } else return null;
```
jesseelizabeth, line 67: nit: can we make this a ternary? `!userLikes.includes(user) ? <ConnectedUserChip key={user} username={user} /> : null`

project/frontend/src/components/SwipeMatchSummary.js
```diff
@@ -0,0 +1,80 @@
...
+
+import ThumbDown from '../assets/thumb_down.svg';
+import ThumbUp from '../assets/thumb_up.svg';
+
+function SwipeMatchSummary(props) {
+  const { restaurants, users, likes } = props;
+
+  const history = useHistory();
+
+  const topPick = restaurants[0];
```
jesseelizabeth, line 18: Will restaurants ever be empty?

jesseelizabeth:
This component looks awesome !

project/frontend/src/components/SwipeMatchSummary.js
alexcostaluiz: Yes it certainly will be empty before they are fetched from the Places API. In that case `topPick` will be undefined so the following if-else will make sure that no null pointers are hit (if `topPick` is undefined we just return `null` for the whole component which I don't think is an issue since this component is only every *visible* at the end of a session when restaurants are certainly populated).

surbhimadan1995: could we just have a return-early check for restaurants.length === 0 in that case? It'd be clearer to read that case
if (empty): return null;
return (<div...); 

(also would reduce the if-block level of nesting as a side effect :D)

project/frontend/src/components/SwipeMatchSummary.js
```diff
@@ -0,0 +1,80 @@
...
+          </div>
+        </div>
+        <table>
+          <tbody>
+            <tr>
+              <td>
+                <img src={ThumbUp} alt='Thumb up' />
+              </td>
+              <td>
+                <label>Likes</label>
```
kpancha, line 45: For all the labels we should have a htmlFor attribute that equals the id of the element it is labeling https://www.w3schools.com/jsref/prop_label_htmlfor.asp

project/frontend/src/components/SwipeMatchSummary.js
```diff
@@ -0,0 +1,81 @@
...
+  const topPick = restaurants[0];
+  const id = topPick.key.id;
+  const name = topPick.key.name;
+  const votes = topPick.fractionLiked * users.length;
+  const userLikes = likes[id];
+
+  return (
+    <div className='swipe-match-summary'>
+      <h4>
+        Your top pick was{' '}
```
miahsanchez, line 31: do we need {' '} ?

project/frontend/src/components/SwipeMatchSummary.js
```diff
@@ -0,0 +1,81 @@
...
+                    !userLikes.includes(user) ? (
+                      <ConnectedUserChip key={user} username={user} />
+                    ) : null
+                  )}
+              </td>
+            </tr>
+          )}
+        </tbody>
+      </table>
+      <button onClick={() => history.replace('/')}>Homepage</button>
```
miahsanchez, line 76: not a fix for now but I think it could be cool to just have an `X` button in the top right corner as opposed to this homepage button. This all looks so great!!!

project/frontend/src/components/SwipeMatchSummary.js
alexcostaluiz: Yea

